PR #33, opened 2025-05-27
[monarch/shipit] Delete monarch_hyperactor/src/bin/process_allocator/…

kiukchung:
D75298941 excludes `**/*_meta.rs` from being sync'ed. The diff that added `main_meta.rs` landed before shipit configerator changes. Deleting `main_meta.rs` to bring ShipIt back to sync. Related task: [T225357761](https://www.internalfb.com/intern/tasks/?t=225357761)


kiukchung:
won't need this since @dcci  is going to reset the repo to do a fresh sync.

PR #98, opened 2025-05-30
Fix cargo resolver warning

mkuchnik:
Summary:
Fix cargo warnings by changing root resolver to `2`.

When installing `monarch` on OSS environments, users may observe warnings for features such as `cargo install --list`:
https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions

Differential Revision: D75716081


dcci (approved):
looks good, thanks

PR #101, opened 2025-05-31
[Packaging] Specify that Monarch is distributed under BSD license

malfet:


setup.py
```diff
@@ -116,6 +116,7 @@ def run(self):
     package_dir={"": "python"},
     python_requires=">= 3.10",
     install_requires=reqs.strip().split("\n"),
+    license="BSD",
```
dcci: We might want to add also

```
license-files = [
    "LICENSE",
]
```

and, can we change the version to 

```
version = "0.0.1"
````

dcci (approved):
LG

PR #174, opened 2025-06-05
build instruction option for non-system nccl

amirafzali:
Differential Revision: D76062785

For build from source, clarifying that conda nccl is an option if the system dep cannot be installed (or if they do not want a system level package). Thanks @mkuchnik  for helping me use your system to debug this


amirafzali:
merged

PR #202, opened 2025-06-09
Add setuptools-rust to requirements and pyproject.toml

dulinriley:
Trying to build fresh from scratch, noticed some dependencies were missing. I also wanted to use python's virtualenv to do the installation work to avoid installing a global torch version, so I added `.venv` to the .gitignore

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,9 @@
-[tool.pytest.ini_options]
+[build-system]
+requires = ["setuptools", "setuptools-rust"]
+build-backend = "setuptools.build_meta"
 
+
+[tool.pytest.ini_options]
 markers = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,5 +1,9 @@
-[tool.pytest.ini_options]
+[build-system]
+requires = ["setuptools", "setuptools-rust"]
```
dulinriley, line 2: This is what the website for setuptools rust suggested: https://pypi.org/project/setuptools-rust/

dulinriley:
This wasn't needed, we had build-requirements.txt for this purpose instead, I didn't see it and assumed "requirements.txt" was the right one

PR #241, opened 2025-06-11
subprocess and parallelize py tests

amirafzali:
Differential Revision: D76448556

- pytest-xdist will spin up a number of processes equal to CPU count and parallelizes the tests for a nice speed increase. it should also create some isolation (previously they all shared a single process context which increases the likelihood of state leaks)
  - let's monitor this and make sure it doesn't spin up hard to debug failures
- also setting a default timeout of 300s. don't think anything should ever take more than 5m...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,5 +64,5 @@ jobs:
         pip install --no-build-isolation .
 
         # Run tests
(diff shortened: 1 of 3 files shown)


amirafzali:
> I wonder if we can also remove the `LC_ALL=C` workaround now. But that can be done separately.

@malfet did we get an upstream issue to track this?

PR #266, opened 2025-06-13
add support for CPU build (no tensor engine), separate build & test steps

amirafzali:


suo:
this is probably called the "no tensor engine" build, as it removes the whole tensor engine functionality, not just cuda

amirafzali:
> this is probably called the "no tensor engine" build, as it removes the whole tensor engine functionality, not just cuda

@suo adjusted the naming but lmk if there's something you prefer

PR #814, opened 2025-08-10
Rust API build

sekyondaMeta:
API docs build update. Setting up the API docs build

### DIFF
--- a/.github/workflows/doc_build.yml
+++ b/.github/workflows/doc_build.yml
@@ -13,63 +13,66 @@ concurrency:
 jobs:
   build:
     name: Build Documentation
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    with:
+      timeout: 120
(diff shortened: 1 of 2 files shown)


.github/workflows/doc_build.yml
```diff
@@ -5,74 +5,70 @@ on:
...
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          submodules: recursive
+  build-and-deploy:
+    name: Build and Deploy Documentation
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    with:
+      timeout: 120
```
johnwhumphreys, line 19: Is 120 seconds always enough for this?  That sounds ambitious.

sekyondaMeta: In my tests, it was building in about 30 mins so I think this is actually 120 minutes

PR #1528, opened 2025-10-14
Use pytest-isolate to make tests more reliable

dulinriley:
Summary:
The pytest-isolate package can support both pytest-timeout and pytest-forked,
to run each test in a separate process.
This should help prevent cross-contamination (one test affects a second test), as well
as preventing tests with fatal signals from taking down the whole test run.

Differential Revision: D84636534

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,4 +5,4 @@ markers = [
 ]
 asyncio_mode = "auto"
 # Default timeout of 5 minutes
-timeout = 300
+isolate_timeout = 300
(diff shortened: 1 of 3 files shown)


dulinriley:
We won't be able to use this at the moment, or other plugins like pytest-forked, with GPU code:
```
RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method
```
We'll either need to change that plugin to use "spawn" instead of "fork", or make our own plugin.
So for now this will not be a good way to get to more reliable pytests.

dulinriley:
I think to fix the aforementioned issue with CUDA I'll just separate tests which use it to have markers, so we run without isolate on those markers

PR #1561, opened 2025-10-16
Bump timeout for stopping actors

dulinriley:
Summary: Let's see if this fixes the problem in test_actor_error

Differential Revision: D84771676


dulinriley:
This didn't fix the problem, will find other solutions.

PR #1652, opened 2025-10-23
rdmaxcel-sys: use ibv_query_device for SGE limit

sielicki:


sielicki:
@dstaay-fb what are your plans for this module, in general? I'm tentatively interested in contributing patches for:

1. alternative hmem support, especially hip/amdgpu support
2. making the mlx5 ibv extension calls optional, depending on whether the nic is in-fact mlx5

but given the performance issues hinted at by this code:

```cpp
        // If we have too many MRs, compact them into a single MR
        if (seg.mrs.size() > max_sge) {
          // TODO: find a safe way to compact with low performance cost.
          // return MAX_SGE error auto err = compact_mrs(pd, seg, access_flags);
          // if (err != 0) {
          //   return err;
          // }
          return RDMAXCEL_MKEY_REG_LIMIT;
        }
```

I'm wondering whether you internally have other plans to rethink this interface with tighter integration with the allocators in pytorch, and whether it's a poor spend of time.

sielicki:
most of the existing dmabuf export code, between nccl/nccl plugins/mpi/ofi/ucx/etc., are using the driver API (`cuMemGetHandleForAddressRange`), but there's an alternative mechanism in cudart to export an entire cudaMemPool_t (`cudaMemPoolExportToShareableHandle`), and I wonder whether that interface can achieve better outcomes for avoiding the problematic unmap/remap thrash and compaction problems. It also avoids some of the portability footguns with using the driver api. Food for thought.

dstaay-fb:
So its probably not obvious from code; but this code path is integrated into CUDA caching allocator.   Idea is when you request an RdmaBuffer for a given tensor, we look into pt allocator and get entire 'segment', register that entire memory range as a MR, and return a view of the MR (base addr + offset).    A second call, is trival - assuming its within existing segment we just return a different offset at rust layer.   In practice segments grow; so we have to create additional MRs but can bind those to the existing mkey via `mlx5dv_wr_mr_list` (so static for life of segment).       This api has a few practical issues' (1) mr's are limited to 2^32-(1 page size), (2) overflow bug in max length (https://github.com/linux-rdma/rdma-core/pull/1650).   But with this patch in practice we can register a 30 * 4GB ~= 120GB segment, which is pretty close to A100 BAR1 limit.     Got any better ideas let me know, but goal is track PT memory allocations, not create/destroy MRs on a per tensor basis.

Re supporting other Accelerators / NIC Vendors, any help there is welcome, in practice we wrote the code with access to NVDA / MLX hardware, but goal is to broaden out with community contributions (folks from AMD have started engagement around HIP/Broadcom NICs) and in theory even AWS EFA should be supportable.

dstaay-fb:
> ```c++
>           // TODO: find a safe way to compact with low performance cost.
>           // return MAX_SGE error auto err = compact_mrs(pd, seg, access_flags);
>           // if (err != 0) {
>           //   return err;
>           // }
> ```

yeah we want avoid this at all costs, probably should drop this code

casteryh:
Can you fix this before we merge it? 
```
raised a warning on
line 164
unused parameter 'max_sge'
Lint code: CLANGTIDY
Lint name: clang-diagnostic-unused-parameter
```

sielicki:
dropped those hunks, sorry about that.

we should probably also replace `MAX_MR_SIZE` with `dev_attr.max_mr_size` and parse `page_size_cap` for alignment, but I am nervous that touching those right now (given the pending patch to rdma-core linked above) will cause issues for you.

PR #1768, opened 2025-11-06
Version guard for AccumulateGrad::add

pzhan9:
Summary: When running `fbpkg build //monarch:monarch_dev` on 3279935a10 we encounter a compile time error that seems to stem from compiling with different versions of PyTorch. We might want to update our build chain. It seems monarch is once compiled against a third part 2.8.0 pytorch and once against the fbcode version. This version guard helps with this version drift if that's intended.

Reviewed By: pzhan9, zdevito, soulitzer

Differential Revision: D86363543

### DIFF
--- a/python/monarch/gradient/_gradient_generator.cpp
+++ b/python/monarch/gradient/_gradient_generator.cpp
@@ -23,6 +23,12 @@
 #include <torch/csrc/autograd/python_function.h> // @manual=//caffe2:torch_extension
 #include <torch/extension.h> // @manual=//caffe2:torch_extension
 
+#define TORCH_VERSION_NEWER_THAN(major, minor, patch)                   \
+  ((TORCH_VERSION_MAJOR > (major)) ||                                   \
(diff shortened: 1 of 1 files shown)


python/monarch/gradient/_gradient_generator.cpp
```diff
@@ -420,12 +426,20 @@ struct GradientGenerator {
     DEBUG_PRINT(
         "// add: " << node->node->name()
                    << ", input_nr=" << static_cast<int>(input_nr) << "\n");
+#if TORCH_VERSION_NEWER_THAN(2, 8, 0)
```
jithunnair-amd, line 429: @pzhan9 Shouldn't this condition have been for 2.9.1? Even the release/2.9 branch has a 4-argument version: https://github.com/pytorch/pytorch/blob/d38164a545b4a4e4e0cf73ce67173f70574890b6/torch/csrc/autograd/input_buffer.h#L31

jithunnair-amd: Addressed by https://github.com/meta-pytorch/monarch/pull/2000

Issue #1774, opened 2025-11-07
ImportError: undefined symbol: _Unwind_GetDataRelBase when importing `monarch` (Rust bindings, Python 3.12, conda)


AlirezaShamsoshoara:
Hi team,
Attempting to `import monarch` fails at runtime with an undefined symbol error originating from the Rust extension `_rust_bindings.so`.
### Repro
Install Monarch from the source on Ubuntu (following the readme) and had issues with that ([1773](https://github.com/meta-pytorch/monarch/issues/1773)).
Solved ([1773](https://github.com/meta-pytorch/monarch/issues/1773)) issue by installing `libunwind` on conda (even though the system has it) it complain w/o the conda one.

Now monarch can be installed but 
Run:
   ```bash
   python -c "import monarch"
```

errors / logs:

```bash
⚡ main ~/monarch python -c "import monarch"
Traceback (most recent call last):
  File "/home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/monarch/__init__.py", line 15, in <module>
    import monarch._rust_bindings  # @manual  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/monarch/_rust_bindings.so: undefined symbol: _Unwind_GetDataRelBase

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/monarch/__init__.py", line 21, in <module>
    import monarch._rust_bindings  # @manual  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/monarch/_rust_bindings.so: undefined symbol: _Unwind_GetDataRelBase
```

### Diagnostics
ldd of the failing extension:

```bash
ldd /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/monarch/_rust_bindings.so
   linux-vdso.so.1 (0x00007ffc59dc1000)
        libunwind.so.8 => /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/monarch/../../../libunwind.so.8 (0x0000774df2ec5000)
[...]

Issue #1773, opened 2025-11-07
Build fails with `rust-lld` undefined references to `lzma_*` via `libunwind-ptrace` when building monarch


AlirezaShamsoshoara:
## Summary
Attempting to build/install the `monarch` Python package (which builds the Rust crate `monarch_hyperactor`) fails at link time with `rust-lld` reporting undefined references to several `lzma_*` symbols. The undefined references originate from `libunwind-ptrace.so`, and linking is rejected due to `--no-allow-shlib-undefined`.

This appears to be a missing or unresolved `liblzma` dependency (XZ) on the link line or a library resolution issue inside a Conda environment.
Observed on commit: `d003266df30c493fb06ab58fd65c220505552b50`
Even though I checked and `liblzma` is installed, still gives me an error

## Steps to Reproduce
Follow the README to build it from the source on **Ubuntu (lightning)** on a CPU machine:

```bash
USE_TENSOR_ENGINE=0 pip install --no-build-isolation .
```

It used to be fine with no error or issue, but recently crashed not sure which commit caused that!

exact error logs:

```bash
-enable-new-dtags" "-Wl,-z,origin" "-Wl,-rpath,$ORIGIN" "-Wl,-rpath,$ORIGIN/.." "-Wl,-rpath,$ORIGIN/../../.." "-Wl,-rpath,/home/zeus/miniconda3/envs/cloudspace/lib"
        = note: some arguments are omitted. use `--verbose` to show all linker arguments
        = note: rust-lld: error: undefined reference: lzma_stream_buffer_decode
                >>> referenced by /usr/lib/gcc/x86_64-linux-gnu/13/../../../x86_64-linux-gnu/libunwind-ptrace.so (disallowed by --no-allow-shlib-undefined)
      
                rust-lld: error: undefined reference: lzma_index_uncompressed_size
                >>> referenced by /usr/lib/gcc/x86_64-linux-gnu/13/../../../x86_64-linux-gnu/libunwind-ptrace.so (disallowed by --no-allow-shlib-undefined)
      
                rust-lld: error: undefined reference: lzma_index_end
                >>> referenced by /usr/lib/gcc/x86_64-linux-gnu/13/../../../x86_64-linux-gnu/libunwind-ptrace.so (disallowed by --no-allow-shlib-undefined)
      
                rust-lld: error: undefined reference: lzma_index_buffer_decode
[...]

PR #1775, opened 2025-11-07
two workaround to build from the source on Ubuntu

AlirezaShamsoshoara:
This PR addresses these two issues (installing monarch from source on Ubuntu and importing monarch):

[1773](https://github.com/meta-pytorch/monarch/issues/1773) 
and
[1774](https://github.com/meta-pytorch/monarch/issues/1774)

There might be another better way to handle these two issues.

### Install libunwind in Conda
To ensure proper stack unwinding and compatibility with certain native libraries, install libunwind from conda-forge:
```bash
# Install the libunwind package from conda-forge
conda install -c conda-forge libunwind
```

### Preload libgcc_s.so.1 from Conda Environment
Some binaries require a specific version of the GCC runtime library (libgcc_s.so.1). To avoid conflicts with system libraries and ensure the correct version is used, preload the library from your active Conda environment:
```bash
# Preload libgcc_s.so.1 from the current Conda environment to ensure correct GCC runtime linking.
# This helps prevent issues with mismatched or missing GCC runtime libraries when running binaries.
export LD_PRELOAD="$CONDA_PREFIX/lib/libgcc_s.so.1:$LD_PRELOAD"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ sudo apt install -y ninja-build
 sudo apt install -y libunwind-dev
 sudo apt install -y clang
 
+# Install the libunwind package on conda
+conda install -c conda-forge libunwind
+
 # Set clang as the default C/C++ compiler
 export CC=clang
 export CXX=clang++
@@ -157,8 +160,12 @@ USE_TENSOR_ENGINE=0 pip install --no-build-isolation .
 # or setup for development
 pip install --no-build-isolation -e .
 
+# Preload libgcc_s.so.1 from the current Conda environment to ensure correct GCC runtime linking,
+export LD_PRELOAD="$CONDA_PREFIX/lib/libgcc_s.so.1:$LD_PRELOAD"
+
 # Verify installation
 pip list | grep monarch
+python -c 'import monarch; print("monarch ok")'
 ```
 
 #### On non-CUDA machines


AlirezaShamsoshoara:
-- cc @pzhan9 @dulinriley @colin2328

dulinriley:
Does https://github.com/meta-pytorch/monarch/pull/1972 help with this issue? It makes libunwind symbols visible at static link time

dawenxi-007:
I had the same issue with python venv #1773. For #1774, not quite sure, looks like the logs published is only a fractional of the total error logs. These issues seem to be related to some missing packages and config. Here was how I got it working for both `USE_TENSOR_ENGINE=0 / 1` in the python venv:

```
sudo apt install -y libunwind-dev  
sudo apt install -y liblzma-dev
```
For `USE_TENSOR_ENGINE=1`, you need the following dependencies as well
```
sudo apt install -y libstdc++-12-dev
sudo apt install -y libnccl2 libnccl-dev 
```
The you need gcc instead of clang as the compiler by the following commands:
```
export CC=gcc  
export CXX=g++
```
Assuming you've installed other required packages, some venv to be set are:
```
export CXXFLAGS="-I/usr/include/c++/11"
export _GLIBCXX_USE_CXX11_ABI=$(python -c 'import torch; print(torch._C._GLIBCXX_USE_CXX11_ABI)')
export CUDA_HOME=/usr/local/cuda
```
Then you need a config for cargo (not sure why it didn't work by venv but requires a dedicated config file here):
```
# Remove any existing .cargo directory first  
rm -rf .cargo

mkdir -p .cargo  
cat > .cargo/config.toml << 'EOF'  
[target.x86_64-unknown-linux-gnu]  
linker = "clang"  
rustflags = [  
    "-C", "link-args=-Wl,--allow-shlib-undefined",  
    "-L", "native=/usr/lib/x86_64-linux-gnu",  
    "-l", "lzma"  
]
env = { BINDGEN_EXTRA_CLANG_ARGS = "-I/usr/include/c++/11 -I/usr/include/x86_64-linux-gnu/c++/11" }
EOF
```

Now in my case, it works for both `USE_TENSOR_ENGINE=0 pip install --no-build-isolation .` and `USE_TENSOR_ENGINE=1 pip install --no-build-isolation .`.

AlirezaShamsoshoara:
Closing this one, as this is not the case anymore with new PyPI stable releases

PR #1830, opened 2025-11-11
Skip some controller tests in OSS

dulinriley:
Summary:
Over a few CI runs we see these tests are occasionally timing out. Put a time bound on
these tests, and skip the ones that often run over the time bound to avoid failing the whole
job.

Differential Revision: D86796388


dulinriley:
This was already deleted in https://github.com/meta-pytorch/monarch/pull/1947, this PR is not needed anymore.

PR #1973, opened 2025-11-21
Test against stable pytorch instead of nightly

dulinriley:
Summary:
Monarch is not using any features from nightly torch that aren't available
in the stable pytorch release.
Using stable can make our builds a little more predictable and avoid version conflicts
unique to nightlies.

Differential Revision: D87667280


amirafzali (approved):

PR #2000, opened 2025-11-25
Update TORCH_VERSION_NEWER_THAN condition

jithunnair-amd:
Even PyTorch 2.9.1 has the 4-argument version of this API: https://github.com/pytorch/pytorch/blob/d38164a545b4a4e4e0cf73ce67173f70574890b6/torch/csrc/autograd/input_buffer.h#L31

Relates to https://github.com/meta-pytorch/monarch/pull/1768/files#r2561710699 

Not sure if the #if block was tested on PyTorch2.9.

PR #2001, opened 2025-11-26
[ROCm][CI] First draft of ROCm build workflow

jithunnair-amd:
This PR introduces a new build workflow for ROCm

* Uses ROCm6.4 PyTorch 2.9.1 wheels to match the bare metal environment on ROCm runners (will switch to PyTorch nightly wheels in a follow-up PR)
* ROCm6.4 almalinux image has devtoolset-14, so need the PATH setting for that since rust needs a new enough compiler
* Set `USE_TENSOR_ENGINE=0` to disable rust errors when building cuda-sys, nccl-sys etc.
* The eventual goal is to invoke this from `ci.yml`. We also enabled the `ciflow/rocm` label-based trigger mechanism for this repo (similar to how we do on pytorch/pytorch), so that we can test this workflow on the PR itself (a `pull_request` trigger will not work to test on the PR because the AWS OIDC step will fail on a forked repo branch).

Depends on https://github.com/meta-pytorch/monarch/pull/2000, https://github.com/meta-pytorch/monarch/pull/2032, and https://github.com/pytorch/test-infra/pull/7650

### DIFF
--- a/.github/workflows/build-rocm.yml
+++ b/.github/workflows/build-rocm.yml
@@ -18,7 +18,7 @@ permissions:
 jobs:
   build-rocm:
     name: Build ROCm (rocm6.4-py3.10)
-    uses: jithunnair-amd/test-infra/.github/workflows/linux_job_v2.yml@patch-5
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
     strategy:
       fail-fast: true
       matrix:


jithunnair-amd:
cc @mreso for awareness

huydhn:
@jithunnair-amd I can see the rocm tag being pushed to https://github.com/meta-pytorch/monarch/tree/ciflow/rocm/2001, so the bot is working as expected.  The reason why it doesn't show up here in your PR is that you need to land the PR first.  For this PR, you can just add `pull_request` trigger to run it, then remove the trigger before landing

malfet:
@huydhn , not that's not the case, at least with the current state of build-rocm.yml workflow, that is missing `on:tag:` dispatch, see https://github.com/jithunnair-amd/monarch/blob/2370a3e877a25649cdef5ac0d1f048acb74178a6/.github/workflows/build-rocm.yml#L4

jithunnair-amd:
> @huydhn , not that's not the case, at least with the current state of build-rocm.yml workflow, that is missing `on:tag:` dispatch, see https://github.com/jithunnair-amd/monarch/blob/2370a3e877a25649cdef5ac0d1f048acb74178a6/.github/workflows/build-rocm.yml#L4

Sorry, it looks like I lost most of my changes (including the `on:tag`) when I tried to rebase the PR branch..... working on fixing it.

jithunnair-amd:
Successfully ran the ROCm build in https://github.com/meta-pytorch/monarch/actions/runs/21033028710/job/60473551338?pr=2001

.github/workflows/build-rocm.yml
```diff
@@ -0,0 +1,52 @@
...
+  build-rocm:
+    name: Build ROCm (rocm6.4-py3.10)
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - name: 4xlargegpu
+            runs-on: linux.rocm.gpu.gfx942.8.meta-pytorch
+            torch-spec: 'torch --index-url https://download.pytorch.org/whl/rocm6.4/'
```
colin2328, line 28: do we not want to use the nightly pre?

jithunnair-amd: @colin2328 As mentioned in the PR description: "will switch to PyTorch nightly wheels in a follow-up PR"

jithunnair-amd:
> there will be a follow up to enable tests in CI?

Yes, we will have a follow-up PR to enable tests. @zstreet87 can probably comment more on this?

Issue #1978, opened 2025-11-24
Build instructions: pip install -r build-requirements.txt overwrites PyTorch nightly installation


grll:
## Description

The README build instructions install PyTorch nightly first, then immediately overwrite it by installing the unversioned `torch` from `build-requirements.txt`.

### Affected sections

- **On Fedora distributions** (lines 105-106)
- **On Ubuntu distributions** (lines 145-146)  
- **On non-CUDA machines** (lines 183-184)

All follow this pattern:
```bash
pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/...
pip install -r build-requirements.txt  # overwrites torch with PyPI stable
```

### Root cause

`build-requirements.txt` contains an unversioned `torch` dependency:
```
torch
setuptools
setuptools-rust
wheel
numpy
```

### Suggested fix

Remove `torch` from `build-requirements.txt` since it's explicitly installed beforehand with specific index URLs. Alternatively, add a comment in the README warning users to skip torch when installing build requirements.

dulinriley:
Thanks for pointing this out, I didn't realize that the extra torch in build-requirements.txt would cause a new one to be fetched.
What we can do is make a `torch-requirements.txt` file and have these contents:
```
--index-url https://download.pytorch.org/whl/nightly/cu128
torch --pre
```
Then remove it from build-requirements. We can't merge the two because `--index-url` controls the whole file, and we wouldn't be able to find the other build requirements.

colin2328:
@dulinriley  @allenwang28   is working on cleaning this up. please work with him

dulinriley:
@allenwang28 I published https://github.com/meta-pytorch/monarch/pull/2025 which will address this. Will this interfere with your changes?

PR #2025, opened 2025-12-02
Add torch-requirements file for the nightly version

dulinriley:
Summary:
Fixes https://github.com/meta-pytorch/monarch/issues/1978

Separate out the torch requirement into a different file, since it requires a different
index. We don't want `build-requirements.txt` to overwrite it.

It is not used in CI because we use "matrix.torch-spec" to change the torch version we
download. We should ensure the requirements file always matches the tested version.

Differential Revision: D87802441


allenwang28 (approved):

PR #2032, opened 2025-12-03
[ROCm][CI] Add ciflow/rocm tag

jithunnair-amd:
This file enables the mechanism of adding a ciflow label on any PR, and the bot creates a corresponding tag on the repo, which can be used as a trigger for a workflow.

### DIFF
--- a/.github/pytorch-probot.yml
+++ b/.github/pytorch-probot.yml
@@ -1,3 +1,2 @@
-ciflow_tracking_issue: 64124
 ciflow_push_tags:
 - ciflow/rocm


jithunnair-amd:
Cc @huydhn

huydhn (approved):

.github/pytorch-probot.yml
```diff
@@ -0,0 +1,3 @@
+ciflow_tracking_issue: 64124
```
huydhn: I don't think this line applies here, so let remove it.  This was used on PyTorch in the early day to provide an opt-in / opt-out mechanism when people use `ciflow`
```suggestion
```

jithunnair-amd:
@mreso Please import this PR so we can get it merged to test PR https://github.com/meta-pytorch/monarch/pull/2001

PR #2083, opened 2025-12-08
Add Kubernetes support via SkyPilot

romilbhardwaj:
Hi Monarch community 👋

This PR adds support for running Monarch workloads on Kubernetes and cloud VMs via [SkyPilot](https://github.com/skypilot-org/skypilot). 

Adds a new `SkyPilotJob` class that provisions Monarch workers on K8s (or any of the 20+ cloud providers supported by SkyPilot):

```python
job = SkyPilotJob(
    meshes={"workers": 2},
    resources=sky.Resources(cloud=sky.Kubernetes(), accelerators="H200:1"), # SkyPilot resources object https://skypilot.readthedocs.io/en/latest/reference/yaml-spec.html#resources
    cluster_name="my-cluster",
    file_mounts={"/local/path": "/worker/path"}, # For any code/data sync
)
```

With this I was able to run the getting started example from Monarch docs on my multi-node Coreweave Kubernetes cluster with H200s. Full example in `examples/skypilot_getting_started.py`.

**Notes:**
- **Cold start:** Currently builds Monarch from source on workers to keep client/worker versions in sync (~7-10 min). A pre-built container image would reduce this to ~30 seconds.
- **Networking:** Assumes workers are in the same network as the client. If client-worker communication is not very chatty, SkyPilot's SSH tunneling could remove this requirement.

---

This is an early PR - I'm looking for feedback on the direction. I'm a maintainer of the SkyPilot project and was trying to run Monarch on my k8s cluster. Figured an integration with SkyPilot would be a good fit.

Happy to iterate on the implementation if this aligns with where you'd like Monarch to go.

### DIFF
--- a/examples/skypilot/.sky.yaml
+++ b/examples/skypilot/.sky.yaml
@@ -1,2 +1,0 @@
-# kubernetes:
-#   remote_identity: NO_UPLOAD


johnwhumphreys:
@romilbhardwaj - I just saw this; sorry for missing it all day.  I'll check it out in the morning and circle back :).

colin2328:
very cool! is this ready for review? (if so push it out of draft state?)

examples/skypilot/skypilot_job.py
```diff
@@ -0,0 +1,488 @@
...
+        file_mounts: Optional[Dict[str, str]] = None,
+    ) -> None:
+        """
+        Args:
+            meshes: Dictionary mapping mesh names to number of nodes.
+                    e.g., {"trainers": 4, "dataloaders": 2}
+            resources: SkyPilot Resources specification for the instances.
+                       If None, uses SkyPilot defaults.
+            cluster_name: Name for the SkyPilot cluster. If None, auto-generated.
+            monarch_port: Port for TCP communication between Monarch workers.
```
colin2328: clarify this is for bootstrap port (its not eh only port for all monarch TCP comm)

python/monarch/_src/job/skypilot.py
```diff
@@ -0,0 +1,488 @@
...
+set -ex
+
+# Add PPA for newer toolchains
+sudo apt-get update
+sudo apt-get install -y software-properties-common
+sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+sudo apt-get update
+
+# Install system dependencies
+sudo apt-get install -y \
```
colin2328: skypilot requires new system level dependencies?

romilbhardwaj: hey @colin2328, these are dependencies for monarch. SkyPilot's dependencies are handled by the SkyPilot runtime during setup. 

They're needed because this PR currently builds Monarch from source on the workers to keep client/worker versions in sync during development. Once this is merged and a Monarch wheel is released to PyPI, the setup simplifies to just `pip install torchmonarch` on workers. 

The current approach is a temporary workaround for the chicken-and-egg problem: SkyPilotJob lives in Monarch source, but there's no released wheel with it yet.

romilbhardwaj: Question on architecture: How should new schedulers extend Monarch's JobTrait? 

Should they:
1. Live in Monarch source (python/monarch/_src/job/)
    * Pro: Clean import path, UX for users (`from monarch.job import SkyPilotJob`)
    * Con: Requires building from source during dev until wheel is released, may need version pinning after release?
2. Live in a separate package (user code)
    * Pro: Don't need to rebuild monarch
    * Con: Different import pattern (`from <userpkg> import SkyPilotJob`?)
3. Something else I'm missing.

Currently I have the SkyPilot implementation in `python/monarch/_src/job/skypilot.py`, but I can refactor it. Any guidance here would be appreciated 🙏

romilbhardwaj:
Thanks @colin2328 @johnwhumphreys - I'll polish the PR and get it ready for review.  Have a question about the architecture [here](https://github.com/meta-pytorch/monarch/pull/2083/files/8511e5a4870ee80214f870aa2f5fa83eafab7efc#r2608349736), happy to do whatever aligns with Monarch's recommended design.

examples/skypilot/skypilot_ddp.py
```diff
@@ -0,0 +1,200 @@
...
+    python skypilot_ddp.py --num-hosts 2 --gpus-per-host 1
+"""
+
+import argparse
+import asyncio
+import logging
+import os
+import sys
+
+# Set timeouts before importing monarch
```
shayne-fletcher: am working on enabling
```python
from monarch.config import configure

configure(
  host_spawn_ready_timeout="300s",
  message_delivery_timeout="300s",
  mesh_proc_spawn_max_idle="300s"
)
```
to replace the use of env-vars here.

romilbhardwaj: Thanks @shayne-fletcher! That'll be super helpful for backends where cold starts can be > 30s.

romilbhardwaj:
Monarch build was slowing down worker launches, so I've refactored along the lines of [option 2](https://github.com/meta-pytorch/monarch/pull/2083#discussion_r2608349736) in https://github.com/meta-pytorch/monarch/pull/2083/commits/20d36e8c08853d9b436582709f281a8c63934eec. Let me know if this is not the right direction :)

Tested on Coreweave k8s cluster with H200s. PR is still a draft, will finish up the examples (torchtitan and ddp) + readme and mark ready for review soon!

colin2328:
@romilbhardwaj  "the author does not have a valid CLA on file." can you acept the CLA? and publish?

examples/skypilot/skypilot_getting_started.py
```diff
@@ -0,0 +1,252 @@
...
+provisioned by SkyPilot (Kubernetes or cloud VMs).
+
+Prerequisites:
+    pip install torchmonarch-nightly
+    pip install skypilot[kubernetes]  # or skypilot[aws], skypilot[gcp], etc.
+    sky check  # Verify SkyPilot configuration
+
+Usage:
+    # Run on Kubernetes:
+    python getting_started.py --cloud kubernetes --num-hosts 2
```
ahmadsharif1: s/getting_started/skypilot_getting_started?

examples/skypilot/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""
```
ahmadsharif1: I will let @colin2328 and @johnwhumphreys chime in as well, but IMO this package should probably go in a subdir here:

https://github.com/meta-pytorch/monarch/tree/49547b120c8902e4383b48072529b85c81187d78/python/monarch/_src/job

I think the package dir structure for independent contributors should have `job/contrib` or `job/providers`

WDYT?

python/monarch/job/__init__.py
```diff
@@ -9,4 +9,11 @@
 from monarch._src.job.slurm import SlurmJob
 
 # Define exports
-__all__ = ["JobTrait", "job_load", "job_loads", "JobState", "LocalJob", "SlurmJob"]
+__all__ = [
```
ahmadsharif1: This seems unrelated to the PR?

examples/skypilot/skypilot_ddp.py
```diff
@@ -0,0 +1,200 @@
...
+from monarch.utils import setup_env_for_distributed
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+# Import SkyPilotJob from local module
+from skypilot_job import SkyPilotJob
+
+try:
+    import sky
+except ImportError:
+    print("ERROR: SkyPilot is not installed. Run: pip install skypilot[kubernetes]")
```
ahmadsharif1: Maybe we should add this as an optional dependency in pyproject.toml if we move skypilot_job.py from examples to the job dir

ahmadsharif1:
I am also wondering how to test this when we move this outside of the examples dir and into the main package.

@romilbhardwaj How does skypilot do end to end testing in its CI? Maybe we should have a minikube/kind test setup in our CI to test this if we expose this to users. We don't want silent breakages if users are going to use this.

examples/skypilot/skypilot_ddp.py
romilbhardwaj: Agreed!

examples/skypilot/__init__.py
romilbhardwaj: Agree, it seems like JobTrait implementations should be in the monarch package. Some more [discussion here](https://github.com/meta-pytorch/monarch/pull/2083#discussion_r2608349736).

We do something similar for community contributed clouds in SkyPilot: they are included as a part of `sky.clouds` [module](https://github.com/skypilot-org/skypilot/tree/master/sky/clouds). 

However for Monarch, it introduces the overhead of build monarch workers from source to keep client/worker versions in sync ([discussion](https://github.com/meta-pytorch/monarch/pull/2083#discussion_r2608342267), [example](https://github.com/meta-pytorch/monarch/blob/8511e5a4870ee80214f870aa2f5fa83eafab7efc/python/monarch/_src/job/skypilot.py#L35-L85) of extra build steps that a worker needs to run). 

Any recommendations on how to add a new module to `monarch.job` while still letting workers use wheels from `torchmonarch-nightly`? Trying to avoid having workers build from src during dev/testing of this PR.

romilbhardwaj:
Cleaned up the readme, added an architecture diagram and included a DDP notebook inspired by the slurm example. 

Tested on coreweave k8s with H200s.

This PR is ready for review now! cc @ahmadsharif1 @colin2328 @johnwhumphreys

romilbhardwaj:
> @romilbhardwaj How does skypilot do end to end testing in its CI? Maybe we should have a minikube/kind test setup in our CI to test this if we expose this to users. We don't want silent breakages if users are going to use this.

Great point. For our release testing, our CI spins up k8s clusters on EKS, GKE, CW and Nebius and runs our smoke test suite which includes model training examples from pytorch, ray, nemo and other popular frameworks.

For lightweight CPU-only tests, some parts of our CI uses kind (via `sky local up`) to create a local k8s cluster. If it is of interest, shouldn't be too hard to edit this example to run on CPUs only for the CI tests.

romilbhardwaj:
Added some flags to allow the example to run on CPU-only k8s clusters

python/monarch/_src/job/skypilot.py
ahmadsharif1: cc: @zdevito as this is an interesting question.

My intuition is that they should live in python/monarch/_src/job even though writing a job or changing the API will requires building from source for testing purposes. The testing cost is only paid by the developer, not the end user.

IIUC, keeping it in a separate package will still require installation on the workers while the code is in development. However it will not require building rust libraries which are probably the main source of the build latency.

It would be nice for JobTrait implementers to not require to build rust binaries while testing their code. I wonder if simply a `cp` of the new files into site-packages dir on the workers would work, though it seems like a hack.

mariusae: Job implementations should be buildable on top of Monarch's public APIs. So, in general, I would favor the latter pattern, which also has good DevEx. I don't think we should accumulate many job implementations in core Monarch.

Note that this is distinct from the question of whether these are part of the Monarch "ecosystem" packages. In other words, we might maintain them in Monarch itself, but build and distribute them as separate packages.

We have previously talked about using pip flavors to bundle such dependencies for this purpose, e.g.,

```
$ pip install torchmonarch[skypilot]
```

etc.

docs/source/examples/README.rst
```diff
@@ -8,6 +8,7 @@ Examples
 - :doc:`distributed_tensors.py <distributed_tensors>`: Shows how to dispatch tensors and tensor level operations to a distributed mesh of workers and GPUs
 - :doc:`debugging.py <debugging>`: Shows how to use the Monarch debugger to debug a distributed program
 - `Multinode Slurm Tutorial <https://docs.pytorch.org/tutorials/intermediate/monarch_distributed_tutorial.html>`_: Multinode distributed training tutorial using Monarch and Slurm to run an SPMD training job.
+- `Multinode Kubernetes examples <https://github.com/pytorch-labs/monarch/tree/main/examples/skypilot>`_: Run Monarch on Kubernetes and cloud VMs via SkyPilot.
```
ahmadsharif1: s/Multinode Kubernetes examples/Running on Kubernetes using Skypilot/g

That way we can distinguish it from native kubernetes integration that may arrive in the future.

docs/source/index.md
```diff
@@ -82,3 +82,4 @@ We welcome contributions from the community! If you're interested in contributin
 - [Demo notebook](https://github.com/meta-pytorch/monarch/blob/main/examples/presentation/presentation.ipynb)
 - [DevX Pytorch tutorial](https://docs.pytorch.org/tutorials/intermediate/monarch_distributed_tutorial.html)
 - [Lightning Monarch blog](https://lightning.ai/meta-ai/environments/large-scale-interactive-training-with-monarch)
+- [Monarch on Kubernetes](https://github.com/meta-pytorch/monarch/tree/main/examples/skypilot)
```
ahmadsharif1: Change this as well to Monarch on Kubernetes using Skypilot (to distinguish it from native Kubernetes integration)

examples/skypilot/__init__.py
ahmadsharif1: Is this resolved now that we have moved the module outside in a separate dir?

examples/skypilot/monarch_quickstart.sky.yaml
```diff
@@ -0,0 +1,90 @@
...
+  echo "=== Installing Python dependencies ==="  
+  uv pip install --system torchmonarch-nightly
+  # Install SkyPilot with Kubernetes support for launching nested clusters
+  uv pip install --system "skypilot[kubernetes]"
+  
+  # Verify installations
+  python -c "import monarch; print(f'Monarch installed: {monarch}')"
+  python -c "import sky; print(f'SkyPilot installed: {sky}')"
+  
+  # Configure SkyPilot to use in-cluster Kubernetes context
```
ahmadsharif1, line 62: I am curious as to what permissions the driver pod has. How can it launch worker pods using `sky`? Where are those permissions configured?

examples/skypilot/monarch_quickstart.sky.yaml
```diff
@@ -0,0 +1,90 @@
...
+  # This allows the driver pod to launch nested SkyPilot clusters
+  unset SKYPILOT_IN_CLUSTER_CONTEXT_NAME
+  sky api start
```
ahmadsharif1, line 65: Is there a way to view the YAML that skypilot itself generates and applies to kubernetes (that contains the permissions, etc.)? Just curious.

examples/skypilot/skypilot_quickstart_driver.py
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/env python3
```
ahmadsharif1, line 1: Sorry about this, but maybe rename this to example_driver.py or something more descriptive?

examples/skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
+"""
+SkyPilotJob for Monarch.
```
ahmadsharif1: Mention that this is JobTrait implementation for SkyPilot

examples/skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+except ImportError:
+    HAS_SKYPILOT = False
+    sky = None  # type: ignore[assignment]
+
+
+logger: logging.Logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+if not logger.handlers:
+    logger.addHandler(logging.StreamHandler(sys.stderr))
+logger.propagate = False
```
ahmadsharif1: Why is this needed? Where logs not working without this?

examples/skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+"""
+DEFAULT_IMAGE_ID = "docker:pytorch/pytorch:2.9.1-cuda12.8-cudnn9-runtime"
+
+
+def _configure_transport() -> None:
+    """Configure the Monarch transport. Deferred import to avoid import errors."""
+    from monarch._rust_bindings.monarch_hyperactor.channel import ChannelTransport
+    from monarch._rust_bindings.monarch_hyperactor.config import configure
+
+    configure(default_transport=ChannelTransport.TcpWithHostname)
```
ahmadsharif1: I am wondering why you needed to use private APIs to do this

examples/skypilot/monarch_skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+        >>> trainers = state.trainers  # HostMesh with 2 nodes
+    """
+
+    def __init__(
+        self,
+        meshes: Dict[str, int],
+        resources: Optional["sky.Resources"] = None,
+        cluster_name: Optional[str] = None,
+        monarch_port: int = DEFAULT_MONARCH_PORT,
+        idle_minutes_to_autostop: Optional[int] = None,
```
ahmadsharif1, line 107: Why is this not seconds instead of minutes?

examples/skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+
+
+logger: logging.Logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+if not logger.handlers:
+    logger.addHandler(logging.StreamHandler(sys.stderr))
+logger.propagate = False
+
+# Default port for Monarch TCP communication
+DEFAULT_MONARCH_PORT = 22222
```
ahmadsharif1: Perhaps this should be called MONARCH_WORKER_PORT

examples/skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+            meshes: Dictionary mapping mesh names to number of nodes.
+                    e.g., {"trainers": 4, "dataloaders": 2}
+            resources: SkyPilot Resources specification for the instances.
+                       If None, uses SkyPilot defaults.
+            cluster_name: Name for the SkyPilot cluster. If None, auto-generated.
+            monarch_port: Port bootstrapping communication between Monarch workers.
+            idle_minutes_to_autostop: If set, cluster will autostop after this
+                                      many minutes of idleness.
+            down_on_autostop: If True, tear down cluster on autostop instead of
+                              just stopping it.
```
ahmadsharif1: Can you describe in kubernetes terms what "stop" and "teardown" mean (in terms of what happens to pods, etc.)?

examples/skypilot/monarch_skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+        cluster_name = self._cluster_name or f"monarch-{os.getpid()}"
+
+        logger.info(f"Launching SkyPilot cluster '{cluster_name}' with {total_nodes} nodes")
+
+        # Launch the cluster
+        try:
+            request_id = sky.launch(
+                task,
+                cluster_name=cluster_name,
+                idle_minutes_to_autostop=self._idle_minutes_to_autostop,
```
ahmadsharif1, line 230: I realize just now that minutes and the `down` bool is part of the sky API so you have to expose it as is... (no easy way of converting it to seconds, for example), so you can ignore the comments above unless the sky API exposes seconds.

examples/skypilot/monarch_skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+                down=self._down_on_autostop,
+            )
+            # Get the result from the request
+            job_id, handle = sky.get(request_id)
+        except Exception as e:
+            logger.error(f"Failed to launch SkyPilot cluster: {e}")
+            raise RuntimeError(f"Failed to launch SkyPilot cluster: {e}") from e
```
ahmadsharif1, line 238: What if line 215 succeeds and line 222 fails? Do we still need to clean that up?

examples/skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+            job_id, handle = sky.get(request_id)
+        except Exception as e:
+            logger.error(f"Failed to launch SkyPilot cluster: {e}")
+            raise RuntimeError(f"Failed to launch SkyPilot cluster: {e}") from e
+
+        self._launched_cluster_name = cluster_name
+        logger.info(f"SkyPilot cluster '{cluster_name}' launched successfully")
+        
+        # Wait for the job to be RUNNING (setup complete, run started)
+        self._wait_for_job_running(cluster_name, job_id, timeout=JOB_TIMEOUT)
```
ahmadsharif1: Again, should you clean something up if this timeout fails if there is state on the kubernetes cluster

examples/skypilot/monarch_skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
...
+'''
+        # Escape single quotes in the Python code for bash
+        escaped_code = python_code.replace("'", "'\"'\"'")
+        # Set timeout env vars
+        env_vars = " ".join([
+            f"export HYPERACTOR_HOST_SPAWN_READY_TIMEOUT={JOB_TIMEOUT}s",
+            f"export HYPERACTOR_MESSAGE_DELIVERY_TIMEOUT={JOB_TIMEOUT}s",
+            f"export HYPERACTOR_MESH_PROC_SPAWN_MAX_IDLE={JOB_TIMEOUT}s",
+        ])
+        return f"{env_vars} && {self._python_exe} -c '{escaped_code}'"
```
ahmadsharif1, line 324: I am wondering if it will be safer to write to a file and run that if we hit commandline length limits...

EDIT: Linux limits are 2MB so it should be okay

examples/skypilot/monarch_skypilot/skypilot_job.py
```diff
@@ -0,0 +1,457 @@
+"""
```
ahmadsharif1, line 1: I am also wondering if this package should be in a subdir outside of the notebooks, etc. I'll leave it to you

examples/skypilot/monarch_quickstart.sky.yaml
romilbhardwaj: By default, SkyPilot automatically creates a service account with minimal permissions it needs and attaches it to the pod. You can find the permissions here: https://docs.skypilot.co/en/latest/cloud-setup/cloud-permissions/kubernetes.html

These permissions also allow SkyPilot running inside the driver pod to create worker pods. 

Optionally, the user can override and specify their own custom service account SkyPilot config.

examples/skypilot/monarch_quickstart.sky.yaml
romilbhardwaj: Yep! If you run `sky launch` with `export SKYPILOT_DEBUG=1`, the terminal output will include a long k8s manifest that will be applied. It's also stored in SkyPilot [internal state db](https://docs.skypilot.co/en/latest/reference/architecture/internals.html#skypilot-state-management).

examples/skypilot/monarch_skypilot/skypilot_job.py
romilbhardwaj: As you correctly noted below, this is inherited from the sky API which uses minutes :)

examples/skypilot/skypilot_job.py
romilbhardwaj: Added a line:
```
                              On Kubernetes, autostop is not 
                              supported and this must be set to True. Pods will
                              be deleted when the SkyPilot cluster is downed.
```

examples/skypilot/monarch_skypilot/skypilot_job.py
romilbhardwaj: I think that'd be a good idea, especially if we later want to support @mariusae's [suggestion](https://github.com/meta-pytorch/monarch/pull/2083#discussion_r2620039432) of using pip flavors like `pip install torchmonarch[skypilot]`

romilbhardwaj:
Thanks for the reviews @ahmadsharif1! I'll wait for a bit for @colin2328's feedback before working through the comments and doing a round of final testing. 

Also I noticed the linter was failing - do I need to take any action? Couldn't find instructions to run the linter myself.

docs/source/index.md
```diff
@@ -82,3 +82,4 @@ We welcome contributions from the community! If you're interested in contributin
 - [Demo notebook](https://github.com/meta-pytorch/monarch/blob/main/examples/presentation/presentation.ipynb)
 - [DevX Pytorch tutorial](https://docs.pytorch.org/tutorials/intermediate/monarch_distributed_tutorial.html)
 - [Lightning Monarch blog](https://lightning.ai/meta-ai/environments/large-scale-interactive-training-with-monarch)
+- [Monarch on Kubernetes using Skypilot](https://github.com/meta-pytorch/monarch/tree/main/examples/skypilot)
```
colin2328: we can drop this, leave I'll just link to the examples in index.rst

colin2328:
@romilbhardwaj , thanks. I'll fix the lints  and push for you. please address the comments first, rest is fine to me.

examples/skypilot/__init__.py
romilbhardwaj: Yes, this is resolved now by moving to `examples/skypilot/monarch_skypilot` module

examples/skypilot/skypilot_quickstart_driver.py
romilbhardwaj: Sounds good, renamed to skypilot_quickstart_driver.py

romilbhardwaj: Sounds good, renamed to skypilot_quickstart_driver.py

examples/skypilot/monarch_skypilot/skypilot_job.py
romilbhardwaj: Good point, added clean up logic both here and L231

examples/skypilot/monarch_skypilot/skypilot_job.py
romilbhardwaj: Moved to `examples/skypilot/monarch_skypilot` for now

romilbhardwaj:
Thanks @colin2328 and @ahmadsharif1 - I think I've addressed all comments. Please lmk if I missed something. This should be ready to go now.

Tested on 2x H200:8 on a Coreweave k8s cluster.

PR #2362, opened 2026-01-22
[ROCm][CI] Build ROCm monarch wheels in ci.yml

jithunnair-amd:
Also, build monarch wheels for ROCm7.1 and ROCm7.2 to align with ROCm versions used for PyTorch nightly wheels...

### DIFF
--- a/.github/workflows/build-rocm.yml
+++ b/.github/workflows/build-rocm.yml
@@ -25,8 +25,8 @@ jobs:
         include:
           - name: "7.1"
             torch-spec: 'torch --extra-index-url https://download.pytorch.org/whl/nightly/rocm7.1/'
-          - name: "7.0"
-            torch-spec: 'torch --extra-index-url https://download.pytorch.org/whl/nightly/rocm7.0/'
+          - name: "7.2"
+            torch-spec: 'torch --extra-index-url https://download.pytorch.org/whl/nightly/rocm7.2/'
     with:
       timeout: 60
       runner: linux.rocm.gpu.gfx942.1.meta-pytorch


.github/workflows/build-rocm.yml
```diff
@@ -17,24 +17,23 @@ permissions:
...
   build-rocm:
-    name: Build ROCm (rocm7.1-py3.10)
+    name: Build ROCm (py3.10)
     uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
     strategy:
       fail-fast: true
       matrix:
         include:
-          - name: 4xlargegpu
-            runs-on: linux.rocm.gpu.gfx942.8.meta-pytorch
```
jithunnair-amd, line 27: Does monarch build require GPUs on the runner? Does it require 8 GPUs on the runner (instead of 1)? @zstreet87 @mreso

zstreet87: I think it only requires 1 GPU but I'll let @mreso confirm

mreso: @jithunnair-amd yes 1 GPU should be fine for the build

jithunnair-amd: Updated runner label to 1-GPU

jithunnair-amd:
All ROCm CI passed. However, since we're working on [bumping PyTorch nightly wheels to ROCm7.2](https://github.com/pytorch/pytorch/pull/174234), we should wait for that so we can use ROCm7.1/7.2 wheels here

.github/workflows/ci.yml
```diff
@@ -19,6 +19,10 @@ jobs:
     name: Build CUDA
     uses: ./.github/workflows/build-cuda.yml
 
+  build-rocm:
```
dulinriley, line 22: If you don't add this to the "Status Check" at the bottom of the job, it may not raise any flags if it gets broken. Is that ok with you that people may land PRs that break ROCm?
I'd recommend that for now as I'm not sure we want to raise the level of ROCm support at the moment.
What will help is if this job is consistently green after many merges, then we'll have confidence that we won't be breaking things.

cc @zdevito and @samlurye who are working on RDMA. If ROCm is built as part of CI, will that complicate RDMA work?

jithunnair-amd: cc @zstreet87 @lizamd

jithunnair-amd: If we do not add the ROCm build job to the status checks, it won't block PRs, but at least will be a clear indicator if any future changes (including RDMA changes) break ROCm builds in any way.

zstreet87: @dulinriley see this PR for my RDMA work https://github.com/meta-pytorch/monarch/pull/2393

samlurye (approved):

jithunnair-amd:
@mreso Can we merge this now?

.github/workflows/ci.yml
dulinriley: Thanks looks like Sam stamped it so I have no objections!

mreso:
Thanks @jithunnair-amd imported and will merge internally, then it gets synced outside

PR #2393, opened 2026-01-27
[ROCM] Rdma rocm clean

zstreet87:
Key Components:

  1. hipify_torch integration - Submodule added for CUDA→HIP conversion
  2. ROCm detection - Automatically detects ROCm installation in build_utils
  3. Custom mappings - .hipify/rdmaxcel_custom_map.json for RDMA-specific conversions
  4. Type consistency - Rust code uses CUDA types (CUdeviceptr) with proper aliasing for ROCm
  5. HIP platform defines - Proper compilation flags for HIP/ROCm builds

### DIFF
--- a/rdmaxcel-sys/src/driver_api.cpp
+++ b/rdmaxcel-sys/src/driver_api.cpp
@@ -12,7 +12,8 @@
 #include <iostream>
 #include <stdexcept>
 
-// Two-level stringify macro to ensure macro arguments are expanded before stringification
+// Two-level stringify macro to ensure macro arguments are expanded before
+// stringification
 #define STRINGIFY_HELPER(x) #x
 #define STRINGIFY(x) STRINGIFY_HELPER(x)
 
@@ -135,11 +136,12 @@ DriverAPI create_driver_api() {
 
   DriverAPI r{};
 
-#define LOOKUP_CUDA_ENTRY(name, sym)                                      \
+#define LOOKUP_CUDA_ENTRY(name, sym)                                           \
   r.name##_ = reinterpret_cast<decltype(&sym)>(dlsym(handle, STRINGIFY(sym))); \
-  if (!r.name##_) {                                                       \
-    throw std::runtime_error(                                             \
-        std::string("[RdmaXcel] Can't find ") + STRINGIFY(sym) + ": " + dlerror()); \
+  if (!r.name##_) {                                                            \
+    throw std::runtime_error(                                                  \
+        std::string("[RdmaXcel] Can't find ") + STRINGIFY(sym) + ": " +        \
+        dlerror());                                                            \
   }
 
   RDMAXCEL_CUDA_DRIVER_API(LOOKUP_CUDA_ENTRY)


build_utils/src/lib.rs
```diff
@@ -429,6 +431,164 @@ pub fn set_python_rpath() {
     }
 }
 
+// =============================================================================
+// ROCm Support
```
samlurye: Please move these ROCm helper functions to their own ROCm-specific file.

.hipify/rdmaxcel_custom_map.json
```diff
@@ -0,0 +1,28 @@
+{
```
samlurye: Just want to echo my and @zdevito's comments from the other PR -- we would both much prefer to have this handled by #ifdef's in the actual code rather than string find-and-replace.

zstreet87: @samlurye made the modifications. This design modifies the code base much more but it is more transparent. Lmk what y'all think and I appreciate it

rdmaxcel-sys/src/driver_api.cpp
```diff
@@ -98,24 +140,38 @@ CUresult rdmaxcel_cuMemGetHandleForAddressRange(
     size_t size,
     CUmemRangeHandleType handleType,
     unsigned long long flags) {
+#ifdef USE_ROCM
+  return rdmaxcel::DriverAPI::get()->hipMemGetHandleForAddressRange_(
```
samlurye: Can you define all of these so that the callable driver method name never actually changes -- just the symbol that gets looked up? So like
```c++
#ifdef USE_ROCM
  #define SYM_MEM_GET_HANDLE_FOR_ADDRESS_RANGE hipMemGetHandleForAddressRange
  #define SYM_MEM_GET_ALLOCATION_GRANULARITY hipMemGetAllocationGranularity
  ...
#else
  #define SYM_MEM_GET_HANDLE_FOR_ADDRESS_RANGE cuMemGetHandleForAddressRange
  #define SYM_MEM_GET_ALLOCATION_GRANULARITY cuMemGetAllocationGranularity
  ...
#endif

#define RDMAXCEL_CUDA_DRIVER_API(_) \
  _(memGetHandleForAddressRange, SYM_MEM_GET_HANDLE_FOR_ADDRESS_RANGE) \
  _(memGetAllocationGranularity, SYM_MEM_GET_ALLOCATION_GRANULARITY) \
  ...
```
Then you can just do
```c++
rdmaxcel::DriverAPI::get()->memGetHandleForAddressRange_()
```
and don't need #ifdef for each call site.

build_utils/src/rocm.rs
```diff
@@ -0,0 +1,175 @@
...
+        .arg("--project-directory")
+        .arg(output_dir)
+        .arg("--v2")
+        .arg("--output-directory")
+        .arg(output_dir);
+
+    // Add custom mapping file if provided
+    if let Some(custom_map) = custom_map_json {
+        cmd.arg("--custom-map-json").arg(custom_map);
+    }
```
samlurye:
```suggestion
```

rdmaxcel-sys/src/rdmaxcel.cpp
```diff
@@ -15,6 +15,15 @@
...
 
+// Platform-specific cast for device pointers
+// In CUDA: CUdeviceptr is unsigned long long, use static_cast from size_t
+// In ROCm: hipDeviceptr_t is void*, use reinterpret_cast from size_t
+#ifdef USE_ROCM
+  #define DEVPTR_CAST(addr) reinterpret_cast<CUdeviceptr>(addr)
+#else
+  #define DEVPTR_CAST(addr) static_cast<CUdeviceptr>(addr)
+#endif
+
```
samlurye, line 28: Since this macro always takes `size_t` and returns `CUdeviceptr`, can you make this a function instead:
```suggestion
CUdeviceptr deviceptr_cast(size_t addr) {
// Platform-specific cast for device pointers
// In CUDA: CUdeviceptr is unsigned long long, use static_cast from size_t
// In ROCm: hipDeviceptr_t is void*, use reinterpret_cast from size_t
#ifdef USE_ROCM
    return reinterpret_cast<CUdeviceptr>(addr);
#else
    return static_cast<CUdeviceptr>(addr);
#endif  
}
```

rdmaxcel-sys/build.rs
```diff
@@ -15,63 +20,95 @@ fn main() {}
...
-            std::process::exit(1);
-        }
-    };
+    // Detect platform: ROCm or CUDA
+    let (is_rocm, compute_home) = detect_platform();
 
-    // Get the directory of the current crate
     let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| {
-        // For buck2 run, we know the package is in fbcode/monarch/rdmaxcel-sys
-        // Get the fbsource directory from the current directory path
```
samlurye, line 44: Please add these comments back.

rdmaxcel-sys/build.rs
```diff
@@ -15,63 +20,95 @@ fn main() {}
...
+    println!("cargo:rerun-if-changed=src/rdmaxcel.cu");
+    println!("cargo:rerun-if-changed=src/driver_api.h");
+    println!("cargo:rerun-if-changed=src/driver_api.cpp");
+
+    // Link against dl for dynamic loading (both platforms)
+    println!("cargo:rustc-link-lib=dl");
+
+    // Platform-specific GPU runtime linking
+    if is_rocm {
+        // ROCm: Link dynamically to HIP runtime
```
samlurye, line 73: Would it be possible to link statically to the HIP runtime for consistency?

rdmaxcel-sys/build.rs
```diff
@@ -113,12 +150,15 @@ fn main() {
         .allowlist_type("poll_context_t")
         .allowlist_type("poll_context")
         .allowlist_type("rdmaxcel_segment_scanner_fn")
+        // Vars
         .allowlist_var("MLX5_.*")
         .allowlist_var("IBV_.*")
-        // Block specific types that are manually defined in lib.rs
+        .allowlist_var("CUDA_SUCCESS")
+        .allowlist_var("CU_.*")
+        // Block manually defined types
```
samlurye: Please revert this comment

rdmaxcel-sys/build.rs
mreso: @samlurye IIRC rocm does not provide a static library that we could use for this

rdmaxcel-sys/build.rs
```diff
@@ -161,174 +191,309 @@ fn main() {
...
+        &manifest_dir,
+        &out_dir,
+        is_rocm,
+    );
+}
 
-                cpp_build.compile("rdmaxcel_cpp");
+/// Detect whether to use ROCm or CUDA
+fn detect_platform() -> (bool, String) {
+    // Try ROCm first (ROCm systems may have CUDA installed too)
```
samlurye: Can we control this with an env var `MONARCH_RDMA_GPU_PLATFORM`? The behavior should be: if only cuda *or* rocm is installed, then just go with that; otherwise, the user needs to set `MONARCH_RDMA_GPU_PLATFORM=[cuda|rocm]`.

samlurye (approved):

PR #2464, opened 2026-01-29
Update cuda versions for all tests and builds on Github

dulinriley:
Summary:
Now that our nightly and stable builds use cuda 12.8, our tests should use it as well.
It shouldn't affect the majority of tests that don't use GPUs, but there are a couple tensor
engine and RDMA tests it could affect.

Reviewed By: allenwang28

Differential Revision: D91807566


allenwang28 (approved):

PR #2476, opened 2026-02-02
Add Kubetorch example for running Monarch from outside cluster

dongreenberg:
## Summary
- Add example showing how to run Monarch on Kubernetes using Kubetorch
- The `KubernetesJob` here is a drop-in replacement for Monarch's `KubernetesJob` that enables:
  1. Using Monarch from outside the cluster via an in-cluster gateway, with no code serialization or networking hoop-jumping. Default connection to the websocket proxy is via kubectl, but users without port-forward access can also use a proper load-balancer by setting KT_API_URL. Unlike connecting to the HostMesh directly, fast persistent connections are not required between the local coordinator and cluster (the "true" coordinator which connects to the hosts is the gateway, which is inside the cluster).
  2. Image/dependency updates in-flight (like a K8s-native version of sync_workspace), including fast differential code sync scalable to thousands of nodes
  3. Programmatic compute provisioning with `kt.Compute`, flexible to easy-mode (specify high-level resources, like cpus, gpus, memory, etc.), advanced controlls (e.g. labels, annotations, tollerations, secrets, volumes, etc.), or BYO manifest (bring your own resource but we provision and manage it). Bringing pre-provisioned compute that you manage is also supported. Dynamic compute provisioning is also supported conveniently from *inside* the cluster with centralized auth via Kubetorch's controller.
  
Note that I had to name the directory "k8s_kubetorch" because it was colliding with the kubetorch python package during Python import inside the cluster.

## Contents
- `README.md` - Comprehensive documentation including architecture, quickstart, and API reference
- `kubetorch_quickstart.py` - Example script demonstrating actor spawning and method calls

## Test plan
- [x] Verify README renders correctly on GitHub
- [x] Run `demo.py` on a cluster with Kubetorch latest (0.4.3) installed


dulinriley:
Sorry hadn't noticed this yet, let me review.
cc @thedavekwon and @ahmadsharif1

examples/k8s_kubetorch/monarch_kubetorch/proxy.py
```diff
@@ -0,0 +1,435 @@
...
+
+        self._resolved = True
+        return self._result
+
+    def result(self, timeout: Optional[float] = None) -> R:
+        """Alias for get() for compatibility."""
+        return self.get(timeout=timeout)
+
+
+class ValueMeshProxy(Generic[R]):
```
dulinriley, line 67: Why do we need a Proxy type for ValueMesh? It seems like it doesn't hold data any differently than a normal value mesh.

examples/k8s_kubetorch/monarch_kubetorch/proxy.py
```diff
@@ -0,0 +1,435 @@
...
+            The result value
+        """
+        if self._resolved:
+            return self._result
+
+        encoded_result = self._gateway.get_future_result(
+            future_id=self._future_id,
+            timeout=timeout,
+        )
+        raw_result = cloudpickle.loads(base64.b64decode(encoded_result))
```
dulinriley, line 51: Why is there a base64 encoding string here?

examples/k8s_kubetorch/monarch_kubetorch/proxy.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def values(self):
+        """Iterate over values."""
+        return self._data.values()
+
+    def __len__(self) -> int:
+        return len(self._data)
+
+
+class EndpointProxy:
```
dulinriley, line 95: Is there a Python class equivalent of the JavaScript "Proxy"? Such that we don't need to define every method again and keep it up-to-date. Could it just have every method call do a `self._gateway.forward(method_name, *args, **kwargs)`?
You can have special cases that check the method name for things that could be handled locally.

Otherwise we'll need to keep updating these types every time there's a change to the core monarch APIs which is unnecessary overhead for people changing monarch

examples/k8s_kubetorch/monarch_kubetorch/kubetorch_job.py
```diff
@@ -0,0 +1,264 @@
...
+        if sync_dirs:
+            dirs_to_sync = [os.path.abspath(d) for d in sync_dirs]
+        else:
+            from kubetorch.resources.callables.utils import locate_working_dir
+
+            working_dir, _, _ = locate_working_dir(os.getcwd())
+            dirs_to_sync = [working_dir]
+
+        for sync_dir in dirs_to_sync:
+            image = image.copy(sync_dir, ".", contents=True)
```
dulinriley, line 132: It seems like these are the only lines related to the code sync of the working directory.
How does `kt.Image.copy` here guarantee the peer-to-peer and scalability guarantees you make in the documentation?

examples/k8s_kubetorch/monarch_kubetorch/gateway.py
```diff
@@ -0,0 +1,413 @@
...
+
+        # Import Monarch and attach to workers
+        from monarch.actor import attach_to_workers, enable_transport
+
+        enable_transport("tcp")
+
+        self._host_mesh = attach_to_workers(
+            ca="trust_all_connections",
+            workers=worker_addresses,
+        )
```
dulinriley, line 163: Could you use the monarch.KubernetesJob object to automate the starting of the workers and attaching to them, assuming this cluster is already pre-provisioned?

cc @thedavekwon

examples/k8s_kubetorch/monarch_kubetorch/gateway.py
```diff
@@ -0,0 +1,413 @@
...
+
+        # Wait for initialization
+        self._host_mesh.initialized.get()
+        time.sleep(5)  # Stabilization delay
```
dulinriley, line 167: What does this wait for?

examples/k8s_kubetorch/monarch_kubetorch/gateway.py
```diff
@@ -0,0 +1,413 @@
...
+    def spawn_procs(
+        self,
+        per_host: Dict[str, int],
+        name: str = "procs",
+    ) -> Dict[str, Any]:
+        """Spawn a ProcMesh on the HostMesh."""
+        if not self._initialized:
+            raise RuntimeError("Gateway not initialized. Call initialize() first.")
+
+        proc_mesh = self._host_mesh.spawn_procs(per_host=per_host, name=name)
```
dulinriley, line 186: What happens if we want multiple host meshes on the same cluster? There doesn't seem to be a way to pass in the HostMesh to be used

examples/k8s_kubetorch/monarch_kubetorch/gateway.py
```diff
@@ -0,0 +1,413 @@
...
+            return True
+
+
+def _get_pod_ip() -> str:
+    """Get the IP address of this pod."""
+    hostname = socket.gethostname()
+    return socket.gethostbyname(hostname)
+
+
+class MonarchGateway:
```
dulinriley, line 47: Couldn't this be an Actor that the client sends messages to?
The main reason to do so is that we have a system of supervision that guarantees certain properties when actors or procs are going down / becoming unreachable, and proxying to another client might defeat most of those nice guarantees.

If we instead have a single actor that allows an exposed port, we get all the benefits of the already-designed system

examples/k8s_kubetorch/README.md
```diff
@@ -0,0 +1,239 @@
...
+# Call on single actor
+single_result = actors.get_value.call_one().get()
+
+# Fire-and-forget broadcast
+actors.reset.broadcast()
+```
+
+## Limitations
+
+Because code serialization is problematic when working across diverse environments, such as between a laptop
```
dulinriley, line 220: Why is this a challenge? Doesn't `cloudpickle` handle most of this complexity?

examples/k8s_kubetorch/README.md
```diff
@@ -0,0 +1,239 @@
...
+## Quickstart
+
+### Prerequisites
+
+1. A Kubernetes cluster with [Kubetorch](https://github.com/kubetorch/kubetorch) installed
+2. `kubectl` configured to access the cluster
+3. Python environment with kubetorch installed: `pip install kubetorch`
+4. Run scripts from this directory (`examples/kubetorch`) so Python can find the `monarch_kubetorch` module
+
+**Note:** `torchmonarch` is only required on the cluster, not locally. Actor classes are imported by name on the cluster, so they must be defined at module level (not inside functions).
```
thedavekwon, line 93: Is there specific reason why users won't isntall torchmonarch locally as well? I understand we don't need it, but it seems to be confusing to write default example to showcase it.

examples/k8s_kubetorch/monarch_kubetorch/gateway.py
thedavekwon: Yeah this should be doable. We also added python-native provisioning that could help implementing this but it requires MonarchMesh CRD and operator to be installed.

PR #2519, opened 2026-02-05
Configure Dockerfiles to build non-root container images for use in Kubernetes

MichaelClifford:
Updates both the Dockerfiles to build non-root container images for use in Kubernetes. Users should no longer need to rely on using the `/tmp` directory for the Kubernetes examples. Instead they can work in `/workspace` directly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,3 +24,13 @@ RUN apt-get update && apt-get install vim -y
 
 # Install kubectl (nice to have on client for debugging; not necessary).
 RUN curl -LO https://dl.k8s.io/release/v1.34.0/bin/linux/amd64/kubectl && chmod +x kubectl && mv kubectl /usr/local/bin
+
+# Set /workspace as the working directory
+WORKDIR /workspace
+
+# Allow non-root user to update /workspace
+RUN chgrp -R 0 /workspace && \
+    chmod -R g=u /workspace
+
+# Set container to run as non-root user
+USER 1001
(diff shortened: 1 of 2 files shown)


dulinriley (changes requested):
Thank you for submitting this PR!

I had some questions about the motivation for this change. I was not familiar with the "user" instruction when writing this Dockerfile, so I read up on it here:
https://www.docker.com/blog/understanding-the-docker-user-instruction/

From what I understand, it's a security concern. If you are running this docker image in production and an attacker breaks out of the application being run, it has root priveleges.

But I'm not sure why this would be important to users of this image. This image doesn't even have a "CMD" that is being run, if you were to deploy this image it wouldn't have any incoming connections at all, or be running anything. It's supposed to be more of a development package that just preinstalls monarch. It's intended that users would do `FROM monarch AS base` and then further customize with their own command to run. And when they do so, they can make their own user access control.

I also don't see much precedent from other images. I based this mostly off the pytorch runtime image, which doesn't specify a user: https://github.com/pytorch/pytorch/blob/main/Dockerfile

Please let me know if there's anything I'm misunderstanding

MichaelClifford:
Thanks for the review @dulinriley! 

This change was motivated by going through the examples in `examples/kubernetes` and seeing that this image is used for deploying the worker and controller pods in a Kubernetes cluster. 

https://github.com/meta-pytorch/monarch/blob/aaa2b4d69d2e21f14f94e207ff6300a96a70af7d/examples/kubernetes/hello_kubernetes_job/manifests/hello_mesh.yaml#L62

In that context, setting a non-root user seemed to be a reasonable addition. But to your point, if "It's intended that users would do `FROM monarch AS base`," perhaps setting the user here is not necessary. In that case, would it make sense to add a Dockerfile and image demonstrating that `FROM monarch AS base` pattern to `/examples/kubernetes/`? If so, I'm  happy to do that and you can close this issue. 

Thanks!

dulinriley:
cc @thedavekwon and @ahmadsharif1 should we update the kubernetes example to pull a different docker container with better security setup? It would be good if there was a way to make a small temporary docker image that doesn't need to be pushed and versioned.

PR #2566, opened 2026-02-10
Change default torch index for MacOS builds

dulinriley:
Summary:
There is no MacOS package for pytorch in the cuda index, so the default "uv sync" on a Mac
will fail. Add some markers to differentiate between the platforms.

Reviewed By: amirafzali

Differential Revision: D92721461


amirafzali (approved):

PR #2638, opened 2026-02-13
Add EFA (Elastic Fabric Adapter) support to RDMA subsystem

cpuhrsch:
Summary

  - Add EFA SRD queue pair support alongside existing mlx5/RC transport
  - EFA device auto-detection via efadv_query_device, with RdmaQpType::Auto selecting EFA vs mlx5dv based on hardware
  - EFA connect (INIT→RTR→RTS + address handle creation) and post operations (write/read via extended verbs, recv via
  ibv_post_recv) implemented in C
  - EFA-specific defaults: gid_index=0, max_sge=1, no RDMA atomics
  - Static linking of libefa.a from rdma-core to avoid conflicts with dynamic libibverbs
  - RC/mlx5 path is unchanged — EFA uses early returns and separate helper methods (connect_via_c, post_op_via_c) to

  Test plan

  - Built from clean clone on EFA-equipped AWS H100 nodes
  - rdma_pingpong.py passes all 4 sizes (10/100/500/1000 MB) × 3 iterations
  - All data and pong verifications PASS
  - Throughput: 2.5–8.0 GB/s across sizes
  - RC/mlx5 path unaffected (existing tests continue to work)

Using [moodist](https://github.com/facebookresearch/moodist)'s approach to EFA via ibverbs (instead of libfabric) significantly reduced the complexity of this.

### DIFF
--- a/examples/rdma_pingpong.py
+++ b/examples/rdma_pingpong.py
@@ -1,0 +1,180 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import mmap
+import os
+import socket
+import sys
+import time
+from typing import Optional
+
+import fire
+import torch
+import xxhash
+from monarch.actor import Actor, endpoint, shutdown_context
+from monarch.rdma import RDMABuffer
+
+
+def _checksum(buf: bytes) -> str:
+    """Compute an xxhash-based checksum hex digest."""
+    return xxhash.xxh64(buf).hexdigest()
+
+
+class PingPongActor(Actor):
+    """Actor that participates in RDMA pingpong."""
+
+    def __init__(self, size_bytes: int, buffer_type: str = "tensor"):
+        self.hostname = socket.gethostname()
+        self.size_bytes = size_bytes
+        self.buffer_type = buffer_type
+
+        if buffer_type == "tensor":
+            n = size_bytes // 4
+            self.data = torch.rand(n, dtype=torch.float32)
+            self.recv_buf = torch.zeros(n, dtype=torch.float32)
+        elif buffer_type == "bytearray":
+            self.data = bytearray(os.urandom(size_bytes))
+            self.recv_buf = bytearray(size_bytes)
+        elif buffer_type == "memoryview":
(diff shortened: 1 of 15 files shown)


cpuhrsch:
CI failures appear to be due to flaky tests? Tests seem unrelated to these changes. 

It seems our CI doesn't have infrastructure to run automated RDMA tests? I see these rdma tests were skipped in CI. However, the pingpong example does work on an AWS instance for what it's worth.

It'd be nice to address the CI concern before landing this.

python/tests/test_rdma.py
```diff
@@ -26,6 +31,80 @@
 )
 
 
+# ---------------------------------------------------------------------------
+# Utility function tests (no hardware needed)
```
cpuhrsch, line 30: These test aren't super powerful and could probably be combined

monarch_rdma/src/rdma_components.rs
```diff
@@ -491,7 +491,11 @@ impl RdmaQueuePair {
     ///
     /// This ensures the hardware has sufficient time to settle after reaching
     /// Ready-to-Send state before the first actual operation.
+    /// Skipped for EFA devices which don't need settling time.
     fn apply_first_op_delay(&self, wr_id: u64) {
```
cpuhrsch, line 496: I think this is kind of weird in general. The need for an explicit delay to have something "settle". But I'm writing this out of ignorance.

examples/rdma_pingpong.py
```diff
@@ -0,0 +1,129 @@
...
+    buf1 = a1.get_buffer.call_one().get()
+    cksum0 = a0.checksum.call_one("data").get()
+    cksum1 = a1.checksum.call_one("data").get()
+
+    print(f"RDMA Pingpong: {data_size_mb} MB x {num_iterations} iters")
+    for i in range(num_iterations):
+        # Ping: a0 reads from a1
+        dt = a0.read_from.call_one(buf1).get()
+        got = a0.checksum.call_one("recv").get()
+        ok = abs(got - cksum1) < abs(cksum1) * 1e-5
```
cpuhrsch: I think this still can be improved. Using an actual hash (like xxhash) would be more appropriate. There's no need to not be precise. Also this script could be extended to try GPU Tensors and CPU memoryviews/bytearrays as well. I see this as a the initial steps towards a more generic rdma remote test script.

PR #2711, opened 2026-02-20
Some improvements to Kubernetes DDP example

andrewsykim:
Some improvements to the Kubernetes DDP example:
* Download individual scripts to remove the need to clone the entire repo 
* Use default namespace. This simplifes many of the steps but also makes it easier for users to override the namespace used in the manifests via `kubectl -n`
* Add docs inline for installing the monarch operator


thedavekwon:
We have been consistently using monarch-tests namespace for other examples in https://github.com/meta-pytorch/monarch/tree/main/examples/kubernetes. Since our manifest already creates namespace for you, it didn't seem that much troublesome. We can change all of them to default namespace if you think that is more friendly as an example. @andrewsykim  WDYT?

PR #2762, opened 2026-02-24
Disable RocM test that is failing until it is fixed

zdevito:
Summary: broken for awhile

Differential Revision: D94253604


jithunnair-amd:
@zdevito Where do you see the ROCm builds being broken for a while? I'm seeing a build from today having passed: https://github.com/meta-pytorch/monarch/actions/runs/22364662660/job/64727308794

Also, this is only the build job, we haven't yet enabled a test job for ROCm afaik.

PR #2710, opened 2026-02-20
[ROCm][CI] Change runner for ROCm build workflow

jithunnair-amd:
`ci.yml` gets [triggered on `pull_request`](https://github.com/meta-pytorch/monarch/blob/b0e32ebe597873196cb02b49ed35fd22eb8d8da6/.github/workflows/ci.yml#L4), which then runs the `build-rocm.yml` workflow on a ROCm runner. However, since the PR branch may not be on `meta-pytorch/monarch`, OIDC login for AWS credentials doesn't work. E.g. [this build](https://github.com/meta-pytorch/monarch/actions/runs/22368457701/job/64740843841?fbclid=IwY2xjawQK6MFleHRuA2FlbQIxMQBicmlkETFCbDBiTjZZWTdvS0hPNHZZc3J0YwZhcHBfaWQBMAABHnuazVWHOFQJoYDDB9wt5JGyLSJYLhadXMJ0ymqCj2GJaE7z-xNeIuUAYMx4_aem_LTi916bH4jh8qmBvxKT7oA)

Also, the build shouldn't need a GPU runner; it just needs a docker image with ROCm installed. Hence this PR switches to a CPU-only runner.

Fixes issue related to https://github.com/meta-pytorch/monarch/pull/2761 and https://github.com/meta-pytorch/monarch/pull/2694

PR #2828, opened 2026-02-27
Re-enable ROCm build workflow triggers

jithunnair-amd:
Companion to https://github.com/meta-pytorch/monarch/pull/2710 to actually enable ROCm CI.


.github/workflows/build-rocm.yml
zstreet87, line 47: This PR should enable USE_TENSOR_ENGINE=1 on ROCm now: https://github.com/meta-pytorch/monarch/pull/2891

PR #2891, opened 2026-03-04
 [ROCM] Add ROCm/HIP Support for RDMA Stack

zstreet87:
This PR adds ROCm support to the RDMA acceleration stack, enabling GPU-direct RDMA on AMD GPUs using RCCL instead of NCCL.

  Summary

  - build_utils: Add ROCm detection (detect_gpu_platform()) and hipify_torch integration for CUDA-to-HIP source conversion
  - nccl-sys: Support RCCL with HIP runtime types, add compatibility layer mapping CUDA names to HIP equivalents
  - rdmaxcel-sys: HIP driver API support for GPU memory registration
  - setup.py: Auto-detect ROCm installations alongside CUDA for Python builds

  Key Design Decisions

  - Build system auto-detects ROCm vs CUDA, or can be explicitly set via MONARCH_RDMA_GPU_PLATFORM=rocm|cuda
  - CUDA-to-HIP conversion uses hipify_torch (added as submodule in deps/hipify_torch)
  - Rust code uses CUDA terminology throughout; ROCm builds include compatibility aliases that map to HIP types
  - Requires ROCm 7.0+

  Testing

  - All 1171 Rust tests pass on ROCm system
  - Python tests pass (groups 1-3 verified)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,8 +142,20 @@ if use_tensor_engine and not torch_config:
     sys.exit(1)
 
 build_tensor_engine = use_tensor_engine and torch_config is not None
-build_cuda = build_tensor_engine and cuda_home is not None
-build_rocm = build_tensor_engine and rocm_home is not None
+
+# GPU platform selection: use MONARCH_RDMA_GPU_PLATFORM env var or auto-detect
+gpu_platform = os.environ.get("MONARCH_RDMA_GPU_PLATFORM", "").lower()
+if gpu_platform and gpu_platform not in ("cuda", "rocm"):
+    sys.exit(f"Invalid MONARCH_RDMA_GPU_PLATFORM={gpu_platform}. Use 'cuda' or 'rocm'")
+if gpu_platform == "rocm" and not rocm_home:
+    sys.exit("MONARCH_RDMA_GPU_PLATFORM=rocm but ROCm not found")
+if gpu_platform == "cuda" and not cuda_home:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -234,7 +262,7 @@ def create_cpp_extension(name: str, sources: List[str]) -> Extension:
 ext_modules = []
 if build_tensor_engine:
     cpp_sources = ["python/monarch/common/init.cpp"]
-    if build_cuda:
+    if build_cuda or build_rocm:
```
samlurye: mock_cuda.cpp definitely isn't compatible with ROCM (and hipifying wouldn't solve it either because it relies on the precise assembly code in the cuda driver). Not to mention it's already broken anyway. So I'd just remove this.

setup.py
```diff
@@ -123,15 +143,21 @@ def get_cuda_home() -> Optional[str]:
...
 build_cuda = build_tensor_engine and cuda_home is not None
+build_rocm = build_tensor_engine and rocm_home is not None
 
 print("=" * 80)
 if build_tensor_engine:
-    print("✓ Building WITH tensor_engine (CUDA/GPU support)")
+    print("✓ Building WITH tensor_engine (GPU support)")
     print(f"  - PyTorch: {torch_config['lib_path']}")
-    print(f"  - CUDA: {cuda_home if build_cuda else 'Not found (CPU-only)'}")
+    if build_cuda:
```
samlurye, line 164: Similar to the rust build, we need to give users a way to choose rocm here if the system has both. Can we just use the same env var?

PR #2907, opened 2026-03-05
Add OTLP MeterProvider for OSS metrics export

thedavekwon:
Summary: Add an OTLP-backed SdkMeterProvider that exports metrics to an OTel Collector via HTTP/protobuf when OTEL_EXPORTER_OTLP_ENDPOINT is set. When the env var is unset, the existing no-op behavior is preserved. The opentelemetry-otlp crate automatically reads standard OTel env vars for endpoint, headers, and timeout configuration.

Differential Revision: D95336575

### DIFF
--- a/hyperactor_telemetry/src/otlp.rs
+++ b/hyperactor_telemetry/src/otlp.rs
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+//! OTLP export for OSS observability.
+//!
+//! When `OTEL_EXPORTER_OTLP_ENDPOINT` is set, this module provides:
+//! - A `SdkMeterProvider` that exports metrics via OTLP/HTTP+protobuf
+//! - An `OtlpTraceSink` that exports traces via OTLP/HTTP+protobuf
+//!
+//! When the env var is unset, both functions return `None`, preserving
+//! the current no-op behavior for OSS builds.
+
+use opentelemetry_sdk::metrics::SdkMeterProvider;
+
+use crate::config::OTEL_METRIC_EXPORT_INTERVAL;
+
+#[allow(dead_code)]
+const OTLP_ENDPOINT_ENV: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
+
+/// Build an OTLP-backed `SdkMeterProvider` if `OTEL_EXPORTER_OTLP_ENDPOINT` is set.
+///
+/// The `opentelemetry-otlp` crate automatically reads standard OTel env vars
+/// (`OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_HEADERS`,
+/// `OTEL_EXPORTER_OTLP_TIMEOUT`, etc.), so callers only need to set those.
+///
+/// Returns `None` if the endpoint is not configured, leaving the global
+/// meter provider as the default no-op.
+#[allow(dead_code)]
+pub fn otlp_meter_provider() -> Option<SdkMeterProvider> {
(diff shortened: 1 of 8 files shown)


hyperactor_telemetry/src/otlp.rs
```diff
@@ -0,0 +1,71 @@
...
+        Ok(e) => e,
+        Err(e) => {
+            eprintln!("[telemetry] failed to build OTLP metric exporter: {}", e);
+            return None;
+        }
+    };
+
+    let interval = hyperactor_config::global::get(OTEL_METRIC_EXPORT_INTERVAL);
+
+    let reader = opentelemetry_sdk::metrics::PeriodicReader::builder(exporter)
```
richabanker, line 52: I'm wondering if Monarch should also add an optional, built-in Prometheus exporter to cater to the large number of Kubernetes users who currently rely on standard Prometheus and PodMonitor workflows. Since a built-in prometheus exporter will allow users who dont currently have an OTel collector setup for their K8s clusters, to integrate Monarch out of the box using their existing Prometheus server via standard scraping.

Another point is, Prometheus scraping (via [PodMonitor, ServiceMonitor](https://prometheus-operator.dev/docs/developer/getting-started/)) automatically handles pod discovery and metadata enrichment, such as namespace and pod_name labels. With OTLP push, users would either have to manually inject this metadata via the Kubernetes Downward API as resource attributes or deploy a collector with the k8sattributes processor to handle the enrichment. Seems like adding more toil for users to add this metadata which they could get for free with Prometheus's PodMonitor.

Since Monarch's instrumentation is built on the OTel SDK, both exporters can be registered on the same MeterProvider. Metrics are recorded once and fanned out to all configured exporters. This will let Prometheus users onboard immediately with no new infrastructure, while OTLP users can push to their backend of choice, and neither group is forced into infra they don't have.

Longer term, the Prometheus exporter can serve as a transitional layer. As the ecosystem converges on OTLP (Kubernetes adopting OTLP natively (not yet panned), Prometheus 2.47+ supports [OTLP ingestion](https://prometheus.io/docs/guides/opentelemetry/)), maybe at that point once this becomes the new reality, we can remove the Prometheus exporter and it would require no other instrumentation changes since we're using OTel SDK from the beginning.

thedavekwon: Hmmm that is good point. I have few concerns regarding using prometheus exporter which is pull based model that differ from our push based model. 

1) IIUC, it launches separate http server for aggregation to support pull based model. This means that every pod will launch http server which can be potentially performance critical. In Otel model, this is just aggregated to how user configures. Let me know if I am misunderstanding it here. 

2) How is PodMonitor usually set up? I am not new to K8s telemetry so I am not familiar with the concept, but is it something like Otel Collector where we can push metrics and let PodMonitor do the aggregation?

cc @dulinriley @johnwhumphreys  interested to hear your thoughts too

richabanker: That's right, prometheus exporter would start a new http server to serve the prom-formatted metrics in a /metrics endpoint but its usually a lightweight endpoint, so performance concerns are probably negligible. The server will only serialize the data when a prom server actually scrapes it (typically done in order of 15-30s) so the overhead is not added to the hot-path of recording metrics but rather at scrape setting of Prometheus.

That said, Prometheus 2.47+ does support OTLP metrics https://prometheus.io/docs/guides/opentelemetry/, so users can push OTLP metrics to Prometheus directly without needing a scrape /metrics endpoint. So another thought is, that Monarch exports OTLP only, and for K8s users that dont have a OTel collector but rather a prometheus pipeline for metrics collection, explicitly document using the Prometheus's OTLP receiver. This avoid the in-process http server while being compatible with existing Prometheus setups, so long as the Prom version is 2.47+ and `--web.enable-otlp-receiver` setting is turned on for it. 

Regarding the PodMonitor, its a purely discovery and pull mechanism, its a K8s CRD from Prometheus Operator that tells Prometheus server which pods to scrape for /metrics. Prometheus finds the pods, scrapes their /metrics endpoint on a schedule and also automatically attach k8s metadata labels like pod_name, namespace_name etc.

PR #2936, opened 2026-03-09
enforce mandatory mTLS for mesh admin HTTP server in meta builds

shayne-fletcher:
Summary:
this diff makes mesh admin mutual-TLS-only in fbcode_build, while preserving the existing best-effort TLS / plain-HTTP fallback in OSS.

it threads client-cert enforcement into try_tls_acceptor, fails admin startup if no usable cert bundle is available at Meta, updates the admin TUI client to reliably load a client identity in the native-tls Buck build, and updates examples, skill docs, and the integration test to use and verify the required mTLS flags.

Differential Revision: D95843188


allenwang28 (approved):

PR #2965, opened 2026-03-10
add attrs field to NodePayload alongside properties

shayne-fletcher:
Summary:
this diff extends NodePayload with an attrs: String field and starts dual-writing actor introspection data into both properties and attrs during the migration. in hyperactor/src/introspect.rs, live_actor_payload now builds attrs through build_actor_attrs, which overlays runtime-owned keys on top of published attrs, normalizes status into status/status_reason, and maps FailureInfo into flat failure_* attrs. it also now explicitly clears stale status_reason and failure_* keys when they are not applicable, so the implementation matches IA-4 and IA-5.

the diff documents the attrs contract on NodePayload with labeled invariants IA-1 through IA-6, including valid-json fallback, dual-write semantics, runtime-key precedence, status/failure shape, and payload totality. all introspection payload construction paths in this change set now populate attrs (using "{}" where no attrs are available), including not-found/error sentinel payloads and mesh/admin payload builders that construct NodePayload directly.

hyperactor/src/actor.rs adds test_ia1_ia4_running_actor_attrs, which validates that live actor payload attrs parse as a JSON object, include a non-prefixed running status, omit status_reason for running actors, and omit failure attrs in the non-failed case. this keeps the change scoped to payload shape and attrs construction; it does not switch downstream consumers from properties to attrs yet.

Differential Revision: D95981241


allenwang28 (approved):

PR #2981, opened 2026-03-11
[monarch] fix subprocess spawning to preserve venv python path

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3005
* #2992
* #2991
* __->__ #2981

`std::env::current_exe()` resolves symlinks, so `.venv/bin/python`
becomes `/usr/local/fbcode/platform010/bin/python3.12`. The resolved
path doesn't find `pyvenv.cfg`, so subprocesses lose the venv's
site-packages and can't import monarch.

Use `argv[0]` instead of `current_exe()` in `run_worker_loop_forever`'s
`BootstrapCommand` construction to preserve the original invocation path.

Differential Revision: [D96176392](https://our.internmc.facebook.com/intern/diff/D96176392/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96176392/)!

PR #2991, opened 2026-03-12
[monarch] add missing build dependencies to README

mariusae:
* #3005
* #2992
* __->__ #2991
* #2981

The rdma-core build requires cmake and ninja-build. Without ninja,
the build falls back to make, which can't build individual file-path
targets and fails with "No rule to make target 'lib/statics/libibverbs.a'".

Added cmake, ninja-build, and protobuf-compiler to the Fedora and
Ubuntu install instructions.

Differential Revision: [D96315289](https://our.internmc.facebook.com/intern/diff/D96315289/)


mariusae:

PR #3008, opened 2026-03-13
Fix SIGABRT when CUDA driver fails to load (#3008)

cpuhrsch:
Summary:

Previously, calling any rdmaxcel C API function would SIGABRT when
`libcuda.so.1` could not be loaded because `DriverAPI::get()` throws
a C++ exception from `dlopen` which propagates across the
`extern "C"` boundary into Rust (UB).

Fix:
1. Add `DriverAPI::try_get() noexcept` to prevent C++ exceptions
   from crossing the FFI boundary.
2. Store the error message and expose it via
   `rdmaxcel_driver_init_error()` so Rust callers get a readable
   error instead of SIGABRT.
3. `is_device_ptr()` returns false when `cuPointerGetAttribute`
   fails for any reason (invalid pointer, no context, etc.), and
   only logs a warning (once) when the driver library itself
   failed to load, allowing CPU-only RDMA to proceed.

If `libcuda.so.1` is not in the default `ld.so` search path, set
`LD_LIBRARY_PATH` to include the directory containing it.

Differential Revision: D96067977


cpuhrsch:
Setting up ghstack

PR #3009, opened 2026-03-13
Buffer messages arriving before CommMeshConfig (#3009)

cpuhrsch:
Summary:

Fix a race condition where messages (CastMessage, ForwardMessage,
ForwardMessageV1) could arrive at a CommActor before its CommMeshConfig
was set, causing `anyhow!("CommMeshConfig has not been set yet")`.

This happens during remotemount when spawn_procs and actor creation
overlap with mesh configuration delivery. The fix buffers pre-config
messages in a `Vec<PendingMessage>` and replays them in order once
CommMeshConfig is received.

Changes:
- Add `PendingMessage` enum with Cast/Forward/ForwardV1 variants
- Add `pending: Vec<PendingMessage>` field to CommActor
- Each handler checks `self.mesh_config.is_none()` and buffers if so
- CommMeshConfig handler replays buffered messages after setting config
- Replace `.ok_or_else()` with `.expect()` since config is guaranteed
  after the buffering check

Differential Revision: D96067983

### DIFF
--- a/hyperactor_mesh/src/comm.rs
+++ b/hyperactor_mesh/src/comm.rs
@@ -116,6 +116,16 @@ pub struct CommActor {
 
     /// The comm actor's mesh configuration.
     mesh_config: Option<CommMeshConfig>,
+
+    /// Messages that arrived before `CommMeshConfig`. Replayed once config is set.
+    pending: Vec<PendingMessage>,
+}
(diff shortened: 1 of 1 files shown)


cpuhrsch:
Setting up ghstack

PR #3011, opened 2026-03-13
Add chunked_fuse: read-only FUSE filesystem for packed data (#3011)

cpuhrsch:
Summary:

Add `chunked_fuse.rs`, a Rust implementation of a read-only FUSE
filesystem backed by pre-packed metadata and byte chunks. Implements
`fuse3::path::PathFilesystem` with handlers for lookup, getattr, read,
readdir, readdirplus, readlink, open, and access.

Key features:
- Zero-copy chunk reads via `Bytes::slice` for single-chunk files
- Multi-chunk assembly for files spanning chunk boundaries
- 1-hour TTL on cached attributes
- Unprivileged mounting via `fusermount3`
- Python bindings: `mount_chunked_fuse()` returns a `FuseMountHandle`
  with `unmount()` method

Reviewed By: zdevito

Differential Revision: D96067978

### DIFF
--- a/monarch_extension/Cargo.toml
+++ b/monarch_extension/Cargo.toml
@@ -17,7 +17,10 @@ crate-type = ["cdylib"]
 anyhow = "1.0.102"
 async-trait = "0.1.86"
 bincode = "1.3.3"
+bytes = { version = "1.11.1", features = ["serde"] }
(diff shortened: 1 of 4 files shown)


cpuhrsch:
Setting up ghstack

PR #3012, opened 2026-03-13
Add parallel TLS transfer actors for remotemount (#3012)

cpuhrsch:
Summary:

Add Rust TLS receiver/sender extensions and Python actor wrappers for
parallel data transfer via persistent TLS connections, bypassing the
hyperactor host_agent relay bottleneck.

**Rust extensions:**
- `tls_receiver.rs`: rustls-based TLS server that accepts parallel
  connections and writes received blocks directly into a caller-provided
  Python buffer (anonymous mmap)
- `tls_sender.rs`: rustls-based TLS client that reads from an mmap'd
  pack file and sends blocks over persistent TLS connections

**Python actors:**
- `TransferShardActor`: receives blocks into a cache-backed mmap via
  `start_persistent_receiver` (TLS listener) and `wait_for_data`
- `SenderShardActor`: connects to a receiver via `connect_persistent`
  and streams blocks via `push_block` / `push_done` over a long-lived
  TLS connection
- `_make_server_ssl_context()` / `_make_client_ssl_context()`: MetaTLS
  certificate helpers for SSLWall-compliant connections
- `classify_workers()`: compares per-block xxhash digests to classify
  workers as fresh, partial, or stale
- cloudpickle registration for pickle-by-value serialization

Each sender-receiver pair gets its own persistent TLS connection,
avoiding repeated ~2s TLS handshake overhead per batch. Blocks are
round-robin distributed across streams using a length-prefixed
protocol: `(offset_u64, size_u64, data)` tuples with a zero-size
sentinel for completion.

Reviewed By: zdevito

Differential Revision: D96067979

### DIFF
--- a/monarch_extension/Cargo.toml
+++ b/monarch_extension/Cargo.toml
@@ -35,6 +35,8 @@ nccl-sys = { path = "../nccl-sys", optional = true }
 ndslice = { version = "0.0.0", path = "../ndslice" }
 pyo3 = { version = "0.26", features = ["anyhow", "multiple-pymethods", "py-clone"] }
 rdmaxcel-sys = { path = "../rdmaxcel-sys", optional = true }
+rustls = "0.23.37"
+rustls-pemfile = "2.2.0"
 serde = { version = "1.0.219", features = ["derive", "rc"] }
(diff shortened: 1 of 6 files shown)


cpuhrsch:
Setting up ghstack

PR #3041, opened 2026-03-16
[hyperactor] add `unsplit` flag to PortRef to prevent port splitting during cast

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* #3050
* __->__ #3041
* #3040
* #3039
* #3038
* #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

\nStack walkthrough: https://www.internalfb.com/intern/phabricator/paste/markdown/P2239132492/
When a message is cast to N actors in a mesh, all PortRef fields are
"split" at each comm actor hop — each hop allocates a local proxy port
that reduces replies before forwarding upstream, creating a hierarchical
reduction tree.

For StreamState subscriptions in the mesh controller, we want all N
ProcAgents to send updates directly to the controller's single subscriber
port. Today this port gets split, creating unnecessary proxy chains. The
`unsplit` flag marks a port as "don't split — broadcast as-is."

API:

    subscriber: this.port().bind().unsplit(),

The flag is added to PortRef, OncePortRef, and UnboundPort so it survives
serialization through the bind/unbind cycle. In split_ports, unsplit ports
are skipped with an early return.

The mesh controller's StreamState subscriber port is marked unsplit so
that all ProcAgents send status updates directly to the controller without
intermediate reduction proxies.

Includes a test (`test_unsplit_port_not_split`) that verifies unsplit
ports are not rewritten during casting and replies arrive directly.

Differential Revision: [D96760758](https://our.internmc.facebook.com/intern/diff/D96760758/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96760758/)!

PR #3040, opened 2026-03-16
[hyperactor_mesh] only allow a single transition to a terminating status

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* #3050
* #3041
* __->__ #3040
* #3039
* #3038
* #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

\nStack walkthrough: https://www.internalfb.com/intern/phabricator/paste/markdown/P2239132492/
Guard `HealthState::maybe_update()` so that once a rank reaches a
terminal status (Stopped or Failed), subsequent updates for the same
rank are rejected regardless of generation. This prevents a stale
streaming update from overwriting a terminal status observed earlier
via polling.

Differential Revision: [D96675192](https://our.internmc.facebook.com/intern/diff/D96675192/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96675192/)!

PR #3039, opened 2026-03-16
[hyperactor_mesh] use StreamState in ActorMeshController for real-time mesh status

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* #3050
* #3041
* #3040
* __->__ #3039
* #3038
* #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

\nStack walkthrough: https://www.internalfb.com/intern/phabricator/paste/markdown/P2239132492/
Subscribe to `StreamState<ActorState>` on all ProcAgents so the mesh
controller receives state changes immediately, complementing the
existing `CheckState` polling loop. Both paths write to
`health_state.statuses` with last-writer-wins resolved by a monotonic
generation counter on each actor state update.

Changes:
- Add `generation` and `timestamp` fields to `resource::State<S>`
- Derive Handler/Bind/Unbind on `State<S>` and Bind/Unbind on `Status`
- Add `impl_bind_unbind_basic` for `Duration` and `SystemTime`
- Track generation in ProcAgent's `ActorInstanceState`, increment on
  spawn, stop, and supervision events
- Add `Handler<State<ActorState>>` to `ActorMeshController`
- Subscribe to `StreamState` in `init()`
- Add `HealthState::maybe_update()` for generation-guarded status updates
- Guard both streaming and polling paths with generation comparison

Differential Revision: [D96760754](https://our.internmc.facebook.com/intern/diff/D96760754/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96760754/)!

PR #3038, opened 2026-03-16
[hyperactor_mesh] add StreamState<S> for streaming actor state subscriptions

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* #3050
* #3041
* #3040
* #3039
* __->__ #3038
* #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

\nStack walkthrough: https://www.internalfb.com/intern/phabricator/paste/markdown/P2239132492/
Add `StreamState<S>` to the resource module — a streaming subscription
that delivers `State<S>` updates whenever the actor's state changes.

ProcAgent stores subscribers per actor in `ActorInstanceState` and
notifies them on stop initiation and supervision events. Dead
subscribers are automatically removed via undeliverable handling
(same pattern as ActorMeshController).

Also:
- Add `ActorInstanceState::to_state()` to consolidate state construction
- Add `ActorInstanceState::notify_subscribers()` for fan-out
- Add `STREAM_STATE_SUBSCRIBER` config attr as a header tag for dead-subscriber cleanup
- Add `handle_undeliverable_message` override in ProcAgent

Differential Revision: [D96760750](https://our.internmc.facebook.com/intern/diff/D96760750/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96760750/)!

PR #3047, opened 2026-03-17
[hyperactor_mesh] add WaitRankStatus for race-free proc mesh shutdown

mariusae:
* __->__ #3047
* #3046
* #3041
* #3040
* #3039
* #3038
* #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

`stop_proc_mesh` sends a stop signal to each proc and then queries status
to confirm procs have stopped. Previously it used `GetRankStatus`, which
replies immediately with the current status. Because the stop signal is
asynchronous, the reply is typically `Stopping` (signal sent, process
hasn't exited yet). `stop_proc_mesh` treated any non-`NotExist` status as
complete and returned, allowing `hosts.shutdown()` to tear down the
HostAgent while `ActorMeshController`s were still polling — causing
undeliverable message errors and crashes.

`WaitRankStatus` fixes this by deferring the reply until the resource's
status reaches a caller-specified threshold. For proc-level status
(HostAgent), this means waiting on the `BootstrapProcHandle`'s watch
channel until the process actually exits. For actor-level status
(ProcAgent), pending waiters are stashed and flushed when supervision
events or stop transitions move the status past the threshold.

The HostAgent handler spawns a background tokio task to avoid blocking
the agent's message loop while waiting.

## Race-free shutdown guarantees

The fix relies on three properties that form a causal chain (this is
all in the "clean shutdown" case):

**1. flush-before-exit:** `ProcAgent::shutdown()` calls `proc.flush()`
before exiting. This drains the outbound message queue, ensuring all
`StreamState<ActorState>{ Stopped }` notifications are handed to the
transport before the process dies.

**2. wait-for-exit-before-proceeding:** `WaitRankStatus` blocks on the
proc's watch channel, which fires only after the process has actually
exited (not just received a stop signal). Combined with (1), when
`stop_proc_mesh` returns, every `StreamState<ActorState>{ Stopped }`
for every actor on every rank is **on the wire**.

**3. stop-before-shutdown ordering:** `worker_procs.stop().get()`
completes before `hosts.shutdown().get()` begins. By this point all
controllers have either already processed the streaming Stopped updates
and stopped polling, or any in-flight polls target dead procs and are
silently dropped (`return_undeliverable(false)`).

### Invariants when `stop_proc_mesh` returns:

- All ProcAgent processes have exited (WaitRankStatus waits on watch
  channel, which fires on exit)
- All `StreamState<ActorState>{ Stopped }` are delivered or are on the
  wire (flush before exit, exit before WaitRankStatus reply)
- Controllers will stop polling (they receive Stopped updates via actor
  in-order delivery, or their polls target dead procs)
- No new actor work is happening (all actor instances have been dropped)

### Invariants when `hosts.shutdown().get()` returns:

- All host processes have exited
- No controller is actively polling
- No in-flight messages to any HostAgent
- All streaming subscriptions are drained

Differential Revision: [D96862238](https://our.internmc.facebook.com/intern/diff/D96862238/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96862238/)!

### DIFF
--- a/hyperactor/src/host.rs
+++ b/hyperactor/src/host.rs
@@ -803,7 +803,7 @@ pub enum LocalProcStatus {
 ///   No OS signals are sent or required.
 pub struct LocalProcManager<S> {
     procs: Arc<Mutex<HashMap<reference::ProcId, Proc>>>,
-    stopping: Arc<Mutex<HashMap<reference::ProcId, LocalProcStatus>>>,
+    stopping: Arc<Mutex<HashMap<reference::ProcId, tokio::sync::watch::Sender<LocalProcStatus>>>>,
     spawn: S,
 }
 
@@ -822,8 +822,8 @@ impl<S> LocalProcManager<S> {
     /// that tears it down.
     ///
     /// Status transitions through `Stopping` -> `Stopped` and is
-    /// observable via [`local_proc_status`]. Idempotent: no-ops if
-    /// the proc is already stopping or stopped.
+    /// observable via [`local_proc_status`] and [`watch`]. Idempotent:
+    /// no-ops if the proc is already stopping or stopped.
     pub async fn request_stop(&self, proc: &reference::ProcId, timeout: Duration, reason: &str) {
         {
             let guard = self.stopping.lock().await;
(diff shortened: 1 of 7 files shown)


mariusae:

PR #3050, opened 2026-03-17
[hyperactor] add message types to undeliverable message errors

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* __->__ #3050
* #3041
* #3040
* #3039
* #3038
* #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

Include the message type name in undeliverable message error output.
Previously these errors only showed sender, dest, headers, and data,
making it hard to identify which message was undeliverable. Now both
`Undeliverable` and `UndeliverableReturned` variants include a
`message type:` line using `envelope.data().typename()`. Also adds
the message type to the assertion message in `monarch_hyperactor`'s
undeliverable handler.

Differential Revision: [D96790834](https://our.internmc.facebook.com/intern/diff/D96790834/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96790834/)!

PR #3037, opened 2026-03-16
[hyperactor] expose flush() on Proc and call it during ProcAgent shutdown

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* #3050
* #3041
* #3040
* #3039
* #3038
* __->__ #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

\nStack walkthrough: https://www.internalfb.com/intern/phabricator/paste/markdown/P2239132492/
Add an inherent `Proc::flush()` method that flushes the forwarder,
ensuring buffered outbound messages (e.g. supervision events) are
wire-delivered before the process exits.

`ProcAgent::shutdown()` is now async and calls `self.proc.flush().await`
before signaling the exit code or calling `process::exit`.

Differential Revision: [D96760753](https://our.internmc.facebook.com/intern/diff/D96760753/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96760753/)!

PR #3036, opened 2026-03-16
[hyperactor_mesh] make ProcAgent stop/stop-all fully async

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* #3050
* #3041
* #3040
* #3039
* #3038
* #3037
* __->__ #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

\nStack walkthrough: https://www.internalfb.com/intern/phabricator/paste/markdown/P2239132492/
Stop and StopAll now only *initiate* the stopping process — they send
stop signals but do not wait or eagerly mark actors as stopped. Actor
status is derived entirely from observed supervision events.

Key changes:
- `ActorInstanceState.stopped` -> `stop_initiated` (tracks intent, not outcome)
- New `supervision_event` field on `ActorInstanceState` replaces the
  ProcAgent-level `supervision_events` HashMap
- `ActorInstanceState::status()` derives Running/Stopping/Stopped/Failed
  from the supervision event and stop_initiated flag
- `StopAll` sends stop signals to all actors, then returns. Shutdown is
  triggered from the supervision event handler once all actors are terminal.
- Remove `destroy_and_wait_except_current` (synchronous shutdown path)

Differential Revision: [D96760752](https://our.internmc.facebook.com/intern/diff/D96760752/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96760752/)!

PR #3035, opened 2026-03-16
[hyperactor_mesh] remove dead StopActor message variant from ProcAgent

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3062
* #3051
* #3050
* #3041
* #3040
* #3039
* #3038
* #3037
* #3036
* __->__ #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

\nStack walkthrough: https://www.internalfb.com/intern/phabricator/paste/markdown/P2239132492/
`StopActor` was never sent as a message — only its macro-generated
`self.stop_actor()` method was called internally. Replace it with a
plain `stop_actor_by_id()` method that sends a stop signal to the proc
without waiting for termination (callers never used the result).

Remove the associated `StopActorResult` type.

Differential Revision: [D96760759](https://our.internmc.facebook.com/intern/diff/D96760759/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96760759/)!

PR #3051, opened 2026-03-17
[hyperactor_mesh] add WaitRankStatus for race-free proc mesh shutdown

mariusae:
* #3062
* __->__ #3051
* #3050
* #3041
* #3040
* #3039
* #3038
* #3037
* #3036
* #3035
* #3034
* #3033
* #3032
* #3031
* #3030
* #3029
* #3028
* #3027
* #3026

`stop_proc_mesh` sends a stop signal to each proc and then queries status
to confirm procs have stopped. Previously it used `GetRankStatus`, which
replies immediately with the current status. Because the stop signal is
asynchronous, the reply is typically `Stopping` (signal sent, process
hasn't exited yet). `stop_proc_mesh` treated any non-`NotExist` status as
complete and returned, allowing `hosts.shutdown()` to tear down the
HostAgent while `ActorMeshController`s were still polling — causing
undeliverable message errors and crashes.

`WaitRankStatus` fixes this by deferring the reply until the resource's
status reaches a caller-specified threshold. For proc-level status
(HostAgent), a `ProcStatusChanged` self-notification pattern routes
watch-channel updates through the actor's message loop: bridge tasks
watch each proc's status channel and send `ProcStatusChanged` to the
actor via a `PortHandle`, which triggers a handler that checks pending
waiters against the current status and replies to any whose threshold
is met. For actor-level status (ProcAgent), pending waiters are stashed
and flushed inline via `notify_status_changed` when supervision events
or stop transitions move the status past the threshold.

Changes by file:
- `proc.rs`: extract `Instance::self_client()` from `self_message_with_delay`
  so background tasks can send to `PortHandle`s without an actor context.
- `host.rs`: add `LocalProcManager::watch()` for subscribing to local
  proc lifecycle changes (mirrors `BootstrapProcManager::watch()`).
- `bootstrap.rs`: add `BootstrapProcManager::watch()`.
- `resource.rs`: add `WaitRankStatus` message type.
- `host_agent.rs`: add `WaitRankStatus` + `ProcStatusChanged` handlers,
  `start_watch_bridge` (one bridge per proc), `notify_proc_status_changed`.
  No `Arc`/`Mutex` — pending waiters are a plain `HashMap` touched only
  from `&mut self` handlers.
- `proc_agent.rs`: add `WaitRankStatus` handler, merge `notify_subscribers`
  and `flush_wait_status` into `notify_status_changed`.
- `host_mesh.rs`: switch `stop_proc_mesh` from `get_rank_status` to
  `wait_rank_status(..., Status::Stopped)`, fix `is_terminating` →
  `is_terminated` in the status check.

## Race-free shutdown guarantees

The fix relies on three properties that form a causal chain (this is
all in the "clean shutdown" case):

**1. flush-before-exit:** `ProcAgent::shutdown()` calls `proc.flush()`
before exiting. This drains the outbound message queue, ensuring all
`StreamState<ActorState>{ Stopped }` notifications are handed to the
transport before the process dies.

**2. wait-for-exit-before-proceeding:** `WaitRankStatus` blocks on the
proc's watch channel, which fires only after the process has actually
exited (not just received a stop signal). Combined with (1), when
`stop_proc_mesh` returns, every `StreamState<ActorState>{ Stopped }`
for every actor on every rank is **on the wire**.

**3. stop-before-shutdown ordering:** `worker_procs.stop().get()`
completes before `hosts.shutdown().get()` begins. By this point all
controllers have either already processed the streaming Stopped updates
and stopped polling, or any in-flight polls target dead procs and are
silently dropped (`return_undeliverable(false)`).

Differential Revision: [D96862238](https://our.internmc.facebook.com/intern/diff/D96862238/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D96862238/)!


mariusae:

PR #3056, opened 2026-03-17
Move Drop on HostMesh to guard object

dulinriley:
Summary:
Previously when a HostMesh is Dropped, it would do a best-effort attempt to
run shutdown. This would make the Hosts unreachable, and also make any procs or
actors unreachable.

This was surprising behavior for users of the Python API, who might do something like this:
```lang=py
def make_procs(self, job: JobTrait) -> ProcMesh:
  hosts = job.state().hosts
  return hosts.spawn_procs(per_host={"gpus": 8})
```

This usage would leave the HostMesh object (and its inner HyHostMesh) unreachable and
will be garbage collected. This would then shutdown the host mesh and the procs would
become unreachable.

Change this so it would only happen when using an `async with` block, which makes it
more clear that resources might not be available after exiting. This doesn't impact
existing uses, they can call shutdown the same way.
After shutdown, all modifying APIs on the HostMesh are unusable (spawn_procs, shutdown, etc.) and will raise a RuntimeError.
The Drop on HostMesh is moved to a transparent guard object that can be used just like
a normal HostMesh.

The main alternative would be to have ProcMesh maintain a reference to the HostMesh which
spawned it, which would fix the immediate problem, but would still have visible side-effects
on `__del__`, which is not desirable.

Differential Revision: D96954833

### DIFF
--- a/hyperactor_mesh/src/host_mesh.rs
+++ b/hyperactor_mesh/src/host_mesh.rs
@@ -25,6 +25,7 @@ pub mod host_agent;
 use std::collections::HashSet;
 use std::hash::Hash;
 use std::ops::Deref;
+use std::ops::DerefMut;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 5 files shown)


dulinriley:
This was merged, not sure why the PR is still open:
https://github.com/meta-pytorch/monarch/commit/5834a728a8258a7056b55c41e4c5740e9c844d6b

PR #3130, opened 2026-03-23
Allow remote procs to bind to the Host router with Proc::attach (#3130)

dulinriley:
Summary:

We want to support the capability of a Proc that lives outside of a Host to attach to
that Host, and piggyback off its routing table and connections.
The main reason to allow this is to bridge network partitions, particularly for an "in-cluster"
set of Hosts and an "out-of-cluster" client Proc.

We need to support 4 message send use cases:
* client Proc -> Procs on Host it attaches to
* client Proc -> Procs on other Hosts in-cluster
* Procs on attached Host -> client Proc
* Procs off attached Host -> client Proc

To do so, we customize the ProcOrDial in the Host. It already handles the case
for messages outbound from a local Proc to a Proc on a different host, we simply
extend this to also check for Procs that are "attached" to the same Host but go to a different
address.

The attach mechanism is a duplex server using MessageEnvelope. Any Proc wanting to attach
simply dials the server, which assigns a new incrementing ProcId called `remote-N`. It binds
this into its routing table. That Proc now gets a new ProcId it can use as its mailbox address.

The duplex is key because it is the only IP+port that needs to be exposed from the cluster.
A client can do `kubectl port-forward 5555:HostA-svc:<duplex port>`, and the bidirectional
nature of the duplex port allows messages to flow backwards out of the cluster back to
localhost. This should keep the network security policies pretty easy. For now we'll attach
to a single Host in the mesh, although there's no requirement for this Host to be part of the
mesh, and we could have a dedicated one in the future.

With this implementation, the 4 send cases work like so:
* client Proc -> Procs on Host:
  send goes through duplex addr -> Host receives and checks destination Proc, sees local, delivers to backend
* client Proc -> Procs on other Hosts in-cluster:
  send goes through duplex addr -> Host receives and checks destination Proc, sees not local, forwards to dialer
* Procs on attached Host -> client Proc:
  send goes through backend addr -> Host receives and checks destination Proc, sees "attached", forwards to duplex addr but going other direction
* Procs off attached Host -> client Proc:
  send goes through frontend addr -> Host receives and checks destination Proc, sees "attached" instead of "local", forwards to duplex addr

TODO: undeliverables need to be sent back to the right place

Differential Revision: D97541960

### DIFF
--- a/hyperactor/src/host.rs
+++ b/hyperactor/src/host.rs
@@ -54,12 +54,18 @@ use std::fmt;
 use std::marker::PhantomData;
 use std::str::FromStr;
 use std::sync::Arc;
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::Ordering;
 use std::time::Duration;
 
 use async_trait::async_trait;
+use dashmap::DashMap;
 use futures::Future;
 use futures::StreamExt;
 use futures::stream;
+use hyperactor_config::Flattrs;
+use serde::Deserialize;
+use serde::Serialize;
 use tokio::process::Child;
 use tokio::process::Command;
 use tokio::sync::Mutex;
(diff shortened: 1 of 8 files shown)


dulinriley:
Going to redo this with ghstack

PR #3153, opened 2026-03-23
Add out-of-cluster support for kubernetes

dulinriley:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #3453
* __->__ #3153

Differential Revision: [D97844187](https://our.internmc.facebook.com/intern/diff/D97844187/)

Using the new `Proc::attach_to_host` API to attach a Proc to another host's mailbox,
we can now support a client that lives externally to the networking environment of other hosts.
This is useful for any environment where there is a cluster with a network partition from the
client interacting with it. In particular, Kubernetes is like this.

The APIs already exist, so we modify KubernetesJob to have a few things:
* A way to specify whether we're in-cluster or out-of-cluster with the KubeConfig class
* A new set of instructions for the example that show how to run in out-of-cluster mode
* When running in the out-of-cluster case, we need to do both:
  * Have at least one of the hosts expose a duplex address.
  We use 0.0.0.0 as the hostname because for the purpose of the example
  the port is exposed via `kubectl port-forward`, which means that the forwarder is running
  on the "localhost" hostname inside that pod
  * Use `context` (or `this_proc` / `this_host` which calls `context`) with an `attach_to`
  that points to the host we use as our proxy. This host is now the client's mailbox

Note that this doesn't work if `context` has already been called directly or transitively.
That is a good area to target for UX improvement, similar to the "enable_transport" problem.
But it should be good enough for now.

TODO: Add duplex port to MonarchMesh CRD type similar to the frontend port?

Test Plan:
Tested on a local kubernetes cluster with kind + podman:
```lang=sh
# builds the distributable for the docker image with local changes
uv build --no-build-isolation --wheel
podman build -d Dockerfile.nightly --format docker \
  -t ghcr.io/meta-pytorch/monarch:local-tag \
  --build-arg MONARCH_WHEELS=dist .
# Save archive to disk to load into kind
podman save ghcr.io/meta-pytorch/monarch:local-tag ../monarch-image

# when using podman, set the environment variable KIND_EXPERIMENTAL_ENGINE=podman
kind create cluster
kind load image-archive ../monarch-image

# setup the pods
kubectl create namespace monarch-tests
kubectl apply -f examples/kubernetes/hello_kubernetes_job/manifests/hello_mesh.yaml

# Now the magic! We can run a monarch script from outside the cluster and access host meshes
uv run --no-build-isolation examples/kubernetes/hello_kubernetes_job/hello_kubernetes_job.py --out-of-cluster
```

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D97844187/)!


examples/kubernetes/hello_kubernetes_job/README.md
```diff
@@ -165,3 +165,94 @@ The `--volcano` flag configures `KubernetesJob` to use Volcano's labels:
...
+## Out-of-Cluster Execution
+
+You can also run the same examples shown here from outside the cluster! The client runs locally
+and sends messages to the mesh inside the cluster. This relies on opening a port
+on a single host in your cluster to use as a proxy for the client's mailbox.
+Make sure at least one of your hosts does `run_worker_loop_forever(..., duplex_address=f"tcp://{hostname}:{port}")`.
+Let's say this port is 34000. You can choose any port you like.
+Make sure that port is exposed. The easiest way is with `kubectl port-forward`:
+```
+kubectl port-forward -n monarch-tests pod/mesh1-0 34000:34000
```
siyuanfoundation: it is generally better to define a service and port-forward to the service. It would make it be a little more convenient to transition to a Gateway if needed.

dulinriley: I see, good idea. I believe we define a service for each monarch mesh host, I'll check if I can use that

PR #3190, opened 2026-03-24
Add ROCm GPU test coverage to CI via matrix-based workflow

zstreet87:
- Add set-matrix.yaml for dynamic CUDA/ROCm build and test matrix selection
  - ROCm-only build-and-test for trigger based on ciflow/rocm label on PRs
  - Both CUDA+ROCm build-and-test for main push and scheduled runs
  - CUDA-only build-and-test for pull_request trigger for all PRs (default)
  - ROCm build for pull_request trigger for PRs without the `ciflow/rocm` label (to ensure that ROCm build jobs run on all PRs, w/ or w/o the `ciflow/rocm` label)
- Replace build-cuda.yml and build-rocm.yml with build-gpu.yml to make name GPU-vendor-agnostic
  - As part of this change, no need to use USE_TENSOR_ENGINE=0 to build for ROCm
- Update test-gpu-python.yml and test-gpu-rust.yml to use matrix
- Add setup_rocm_environment() to common-setup.sh
- Update ci.yml to trigger on `ciflow/rocm` tag push (as a result of `ciflow/rocm` label)
- Update ci.yml final status check logic

### DIFF
--- a/rdmaxcel-sys/src/lib.rs
+++ b/rdmaxcel-sys/src/lib.rs
@@ -54,6 +54,8 @@ mod inner {
             HIP_POINTER_ATTRIBUTE_MEMORY_TYPE;
         pub const CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL: hipPointer_attribute =
             HIP_POINTER_ATTRIBUTE_DEVICE_ORDINAL;
+        pub const CU_POINTER_ATTRIBUTE_CONTEXT: hipPointer_attribute =
+            HIP_POINTER_ATTRIBUTE_CONTEXT;
 
         // Memory types
         pub const CU_MEMORYTYPE_DEVICE: u32 = 2; // hipMemoryTypeDevice = 2


.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,105 @@
...
+      matrix: ${{ steps.set.outputs.matrix }}
+    env:
+      # Event flags evaluated by github actions before the step runs:
+      IS_MAIN_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+      IS_SCHEDULE:  ${{ github.event_name == 'schedule' }}
+      IS_ROCM_TAG:  ${{ startsWith(github.ref, 'refs/tags/ciflow/rocm/') }}
+      TRIGGERED_ROCM_LABEL: ${{ github.event_name == 'pull_request' && github.event.action == 'labeled' }}
+
+      # Default CUDA & ROCm runners
+      DEFAULT_ROCM_RUNNER: linux.rocm.gpu.gfx942.8
```
mreso: IIUC the .8 means that this node will provide 8 gpus. Did I get this right? Do we need that many? linux.g5.4xlarge.nvidia.gpu provides a single A10 if I am not mistaken.

jithunnair-amd: Agree. Let's not use 8-GPU runner if we don't need them; use the min GPUs needed

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+          # Use default value as 'false' for unset environment variables
+          IS_MAIN_PUSH="${IS_MAIN_PUSH:-false}"
+          IS_SCHEDULE="${IS_SCHEDULE:-false}"
+          IS_ROCM_TAG="${IS_ROCM_TAG:-false}"
+          TRIGGERED_ROCM_LABEL="${TRIGGERED_ROCM_LABEL:-false}"
+
+          # Decide which matrix entries to include based on event type
+          # Runs ROCm only for push tag OR when PR label gets triggered
+          if [[ "$IS_ROCM_TAG" == "true" || "$TRIGGERED_ROCM_LABEL" == "true" ]]; then
```
jithunnair-amd: @zstreet87 IIUC, this will include the ROCm matrix if it's the `ciflow/rocm` tag, OR any label is added to a PR? Why the latter?

.github/workflows/ci.yml
```diff
@@ -36,10 +36,11 @@ jobs:
 
   test-gpu-python:
     name: Test GPU Python
-    needs: build-cuda
+    needs: [build-cuda, build-rocm]
     uses: ./.github/workflows/test-gpu-python.yml
     with:
-      artifact-name: monarch-cuda-${{ github.sha }}-py3.10
+      artifact-name-cuda: monarch-cuda-${{ github.sha }}-py3.10
+      artifact-name-rocm: monarch-rocm7.2-${{ github.sha }}
```
jithunnair-amd: Suggest not including the ROCm version in the artifact name to reduce maintenance burden and be consistent with CUDA artifact naming

```suggestion
      artifact-name-rocm: monarch-rocm-${{ github.sha }}-py3.10
```

.github/workflows/ci.yml
```diff
@@ -54,10 +55,11 @@ jobs:
 
   test-gpu-rust:
     name: Test GPU Rust
-    needs: build-cuda
+    needs: [build-cuda, build-rocm]
     uses: ./.github/workflows/test-gpu-rust.yml
     with:
-      artifact-name: monarch-cuda-${{ github.sha }}-py3.10
+      artifact-name-cuda: monarch-cuda-${{ github.sha }}-py3.10
+      artifact-name-rocm: monarch-rocm7.2-${{ github.sha }}
```
jithunnair-amd: Suggest not including the ROCm version in the artifact name to reduce maintenance burden and be consistent with CUDA artifact naming

```suggestion
      artifact-name-rocm: monarch-rocm-${{ github.sha }}-py3.10
```

.github/workflows/test-gpu-rust.yml
```diff
@@ -3,35 +3,38 @@ name: Test GPU Rust
...
+      matrix: ${{ fromJSON(needs.set-matrix.outputs.matrix) }}
     with:
       timeout: 120
-      runner: ${{ matrix.runs-on }}
+      runner: ${{ matrix.runner }}
       gpu-arch-type: ${{ matrix.gpu-arch-type }}
       gpu-arch-version: ${{ matrix.gpu-arch-version }}
       submodules: recursive
-      download-artifact: ${{ inputs.artifact-name }}
+      download-artifact: ${{ matrix.name == 'cuda' && inputs.artifact-name-cuda || inputs.artifact-name-rocm }}
```
jithunnair-amd:
```suggestion
      download-artifact: monarch-${{ matrix.name }}-${{ github.sha }}-py3.10
```

.github/workflows/test-gpu-rust.yml
```diff
@@ -3,35 +3,38 @@ name: Test GPU Rust
...
-        description: 'Wheel artifact name from build workflow'
+      artifact-name-cuda:
+        description: 'CUDA wheel artifact name'
         required: true
         type: string
+      artifact-name-rocm:
+        description: 'ROCm wheel artifact name'
+        required: false
+        type: string
+        default: ''
```
jithunnair-amd: This whole `inputs` section can be deleted since set-matrix.yaml can construct the artifact names

.github/workflows/test-gpu-python.yml
```diff
@@ -3,35 +3,38 @@ name: Test GPU Python
...
-        description: 'Wheel artifact name from build workflow'
+      artifact-name-cuda:
+        description: 'CUDA wheel artifact name'
         required: true
         type: string
+      artifact-name-rocm:
+        description: 'ROCm wheel artifact name'
+        required: false
+        type: string
+        default: ''
```
jithunnair-amd: This whole inputs section can be deleted since set-matrix.yaml can construct the artifact names

.github/workflows/test-gpu-python.yml
```diff
@@ -3,35 +3,38 @@ name: Test GPU Python
...
+      matrix: ${{ fromJSON(needs.set-matrix.outputs.matrix) }}
     with:
       timeout: 120
-      runner: ${{ matrix.runs-on }}
+      runner: ${{ matrix.runner }}
       gpu-arch-type: ${{ matrix.gpu-arch-type }}
       gpu-arch-version: ${{ matrix.gpu-arch-version }}
       submodules: recursive
-      download-artifact: ${{ inputs.artifact-name }}
+      download-artifact: ${{ matrix.name == 'cuda' && inputs.artifact-name-cuda || inputs.artifact-name-rocm }}
```
jithunnair-amd:
```suggestion
      download-artifact: monarch-${{ matrix.name }}-${{ github.sha }}-py3.10
```

.github/workflows/set-matrix.yaml
jithunnair-amd: Also, since `ci.yml` doesn't have any triggers based on push to tags, I don't think `IS_ROCM_TAG` would be `True` for any of the runs?

Keeping ROCm CI capacity constraints in mind, we should only trigger on PRs if the `ciflow/rocm` label is applied, unless the ROCm tests run in a short (<15min) time.

.github/workflows/set-matrix.yaml
jithunnair-amd: Updated to 1-GPU runner label

.github/workflows/set-matrix.yaml
jithunnair-amd: In fact, let's remove `"$TRIGGERED_ROCM_LABEL" == "true"`, because: 1) we're never going to trigger on the `github.event.action == 'labeled'` unless explicitly specified in the trigger clause in `ci.yml`, and 2) we shouldn't need it because the tagging that happens as a result of the label, will ensure that `"$IS_ROCM_TAG" == "true"`, and 3) the tag-based trigger will also ensure that *every* update to a PR labeled with `ciflow/rocm` will trigger the workflow.

By the same token, let's add a tag-based trigger for ciflow/rocm to ci.yml.

.github/workflows/ci.yml
jithunnair-amd, line 10:
```suggestion
push:
  branches:
    - main
  tags:
    - ciflow/rocm/*
```

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,95 @@
...
+        value: ${{ jobs.set.outputs.matrix }}
+
+jobs:
+  set:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set.outputs.matrix }}
+    env:
+      # Event flags evaluated by github actions before the step runs:
+      IS_MAIN_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
jithunnair-amd: I don't think we need to have the `github.ref` check here, because that duplicates the ref specification in `ci.yml`, unless we really want to only restrict ROCm to `main`. For eg. if, at some point, ci.yml also adds `release/*` in the `branches` section, we would also want to trigger ROCm CI on pushes to those branches. Have a `github.ref == 'refs/heads/main'` here will silently disabled ROCm CI on those branches.
```suggestion
      IS_MAIN_PUSH: ${{ github.event_name == 'push' }}
```

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,95 @@
...
+            "torch-spec": "--pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu128"
+          }
+          EOF
+          )
+
+          # Use default value as 'false' for unset environment variables
+          IS_MAIN_PUSH="${IS_MAIN_PUSH:-false}"
+          IS_SCHEDULE="${IS_SCHEDULE:-false}"
+          IS_ROCM_TAG="${IS_ROCM_TAG:-false}"
+          TRIGGERED_ROCM_LABEL="${TRIGGERED_ROCM_LABEL:-false}"
```
jithunnair-amd: Remove as per https://github.com/meta-pytorch/monarch/pull/3190/changes#r3002474941, since not needed
```suggestion
```

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,95 @@
...
+  set:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set.outputs.matrix }}
+    env:
+      # Event flags evaluated by github actions before the step runs:
+      IS_MAIN_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+      IS_SCHEDULE:  ${{ github.event_name == 'schedule' }}
+      IS_ROCM_TAG:  ${{ startsWith(github.ref, 'refs/tags/ciflow/rocm/') }}
+      TRIGGERED_ROCM_LABEL: ${{ github.event_name == 'pull_request' && github.event.action == 'labeled' }}
```
jithunnair-amd: Remove as per https://github.com/meta-pytorch/monarch/pull/3190/changes#r3002474941, since not needed
```suggestion
```

.github/workflows/set-matrix.yaml
jithunnair-amd:
```suggestion
          if [[ "$IS_ROCM_TAG" == "true" ]]; then
```

.github/workflows/build-gpu.yml
```diff
@@ -0,0 +1,37 @@
...
+  build:
+    name: Build ${{ matrix.name }} (py${{ matrix.python-version }})
+    needs: set-matrix
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.set-matrix.outputs.build-matrix) }}
+    with:
+      timeout: 60
+      runner: linux.c7i.2xlarge
```
dulinriley, line 27: Why not `matrix.runs-on` like before?
Also, build-cuda used to run on `linux.g5.4xlarge.nvidia.gpu`. Don't we still need that here?

.github/workflows/build-rocm.yml
```diff
@@ -1,48 +0,0 @@
...
-        # Setup build environment (conda + system deps + rust + build deps)
-        setup_build_environment
-        # Install torch nightly
-        pip install ${{ matrix.torch-spec }}
-        pip install -r build-requirements.txt
-        # Setup Tensor Engine
-        setup_tensor_engine
-        # Build monarch (ROCm version)
-        # TODO TEMPORARY: Use USE_TENSOR_ENGINE=0 to avoid Rust build errors with cuda-sys, nccl-sys etc.
-        USE_TENSOR_ENGINE=0 python setup.py bdist_wheel
```
dulinriley, line 48: Do you not need this in build-gpu anymore? Is there some other way that cuda-sys knows when we're trying to install for a ROCm environment?

.github/workflows/ci.yml
```diff
@@ -8,63 +8,62 @@ on:
   push:
     branches:
       - main
+    tags:
+      - ciflow/rocm/*
```
dulinriley, line 12: What's up with these tags? I thought your summary mentioned "labels" on PRs instead?

.github/workflows/ci.yml
```diff
@@ -8,63 +8,62 @@ on:
...
-    uses: ./.github/workflows/build-cuda.yml
-
-  build-rocm:
-    name: Build ROCm
-    uses: ./.github/workflows/build-rocm.yml
+  build-gpu:
+    name: Build GPU
+    uses: ./.github/workflows/build-gpu.yml
 
+  # CPU jobs skip on ROCm tags since they already run in the PR workflow
```
dulinriley: CPU jobs should run on main and in PRs. The difference being that the PR merge base may be different than where it lands on main at push time.
So I'd recommend always running a CPU job, even if it is a ROCm-tagged commit

.github/workflows/ci.yml
```diff
@@ -8,63 +8,62 @@ on:
...
     with:
       artifact-name: monarch-cpu-${{ github.sha }}-py3.10
 
   test-gpu-python:
     name: Test GPU Python
-    needs: build-cuda
+    needs: build-gpu
     uses: ./.github/workflows/test-gpu-python.yml
-    with:
-      artifact-name: monarch-cuda-${{ github.sha }}-py3.10
```
dulinriley, line 44: Why remove the artifact-name here?

.github/workflows/ci.yml
```diff
@@ -74,14 +73,5 @@ jobs:
...
-             [[ "${{ needs.test-gpu-python.result }}" != "success" ]] ||
-             [[ "${{ needs.test-cpu-rust.result }}" != "success" ]] ||
-             [[ "${{ needs.test-cpu-rust-arm64.result }}" != "success" ]] ||
-             [[ "${{ needs.test-gpu-rust.result }}" != "success" ]]; then
-            echo "One or more jobs failed"
-            exit 1
-          else
-            echo "All jobs passed"
-          fi
+        if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
```
dulinriley, line 84: Nice simplification, didn't know about this feature

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,84 @@
...
+      runner-cuda:
+        description: 'Override default CUDA runner label'
+        required: false
+        type: string
+        default: linux.g5.4xlarge.nvidia.gpu
+    outputs:
+      matrix:
+        description: Test matrix (py3.10 only)
+        value: ${{ jobs.set.outputs.matrix }}
+      build-matrix:
```
dulinriley, line 20: These names are not very descriptive. You should call this one `all-python-versions-matrix` instead or something

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+          # Define build matrices (includes python-version)
+          CUDA_PY310='{"name": "cuda", "python-version": "3.10", "docker-image": "pytorch/manylinux2_28-builder:cuda12.8", "torch-spec": "--pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu128"}'
+          CUDA_PY312='{"name": "cuda", "python-version": "3.12", "docker-image": "pytorch/manylinux2_28-builder:cuda12.8", "torch-spec": "--pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu128"}'
+          ROCM_71_PY310='{"name": "rocm7.1", "python-version": "3.10", "docker-image": "pytorch/almalinux-builder:rocm7.1", "torch-spec": "torch --extra-index-url https://download.pytorch.org/whl/nightly/rocm7.1/"}'
+          ROCM_72_PY310='{"name": "rocm7.2", "python-version": "3.10", "docker-image": "pytorch/almalinux-builder:rocm7.2", "torch-spec": "torch --extra-index-url https://download.pytorch.org/whl/nightly/rocm7.2/"}'
+
+          # Build matrix logic:
+          # - ROCm tag: ROCm only (CUDA already built in PR run)
```
dulinriley, line 53: Build both on every run, especially because if you tagged a commit as specifically changing things about ROCm, I bet there's a higher likelihood on average that it breaks something about our GPU builds with CUDA (compared to, say, a python-only or CPU-only change)

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,84 @@
...
+            BUILD_MATRIX="[$ROCM_71_PY310, $ROCM_72_PY310]"
+          elif [[ "$IS_PUSH" == "true" || "$IS_SCHEDULE" == "true" ]]; then
+            BUILD_MATRIX="[$CUDA_PY310, $CUDA_PY312, $ROCM_71_PY310, $ROCM_72_PY310]"
+          elif [[ "$HAS_ROCM_LABEL" == "true" ]]; then
+            BUILD_MATRIX="[$CUDA_PY310, $CUDA_PY312]"
+          else
+            BUILD_MATRIX="[$CUDA_PY310, $CUDA_PY312, $ROCM_71_PY310, $ROCM_72_PY310]"
+          fi
+
+          # Test matrix: ROCm tested only on push/schedule/tag
```
dulinriley, line 67: This one I agree with for now, because until we have confidence in the stability of this CI job, we don't want to block PRs from merging that may or may not be breaking it

.github/workflows/test-gpu-python.yml
```diff
@@ -2,36 +2,29 @@ name: Test GPU Python
...
-            torch-spec: '--pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu128'
-            gpu-arch-type: "cuda"
-            gpu-arch-version: "12.8"
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.set-matrix.outputs.matrix) }}
     with:
       timeout: 120
-      runner: ${{ matrix.runs-on }}
+      runner: ${{ matrix.runner }}
       gpu-arch-type: ${{ matrix.gpu-arch-type }}
```
dulinriley, line 24: Do you need a different `dockerImage` here as well for rocm test-gpu-python?

.github/workflows/ci.yml
```diff
@@ -8,63 +8,62 @@ on:
...
-      artifact-name: monarch-cuda-${{ github.sha }}-py3.10
 
   build-docker:
     name: Build Docker image
-    needs: build-cuda
+    needs: build-gpu
+    if: ${{ !startsWith(github.ref, 'refs/tags/ciflow/rocm/') }}
     uses: ./.github/workflows/build-docker.yml
     with:
-      # Docker image requires python 3.12
```
jithunnair-amd, line 69: @zstreet87 why are we removing this comment? Is it not correct?

.github/workflows/build-gpu.yml
jithunnair-amd: > Why not matrix.runs-on like before? Also, build-cuda used to run on linux.g5.4xlarge.nvidia.gpu. Don't we still need that here?

The build matrix [doesn't have](https://github.com/meta-pytorch/monarch/pull/3190/changes#diff-5e89c9711a76fb25b73cde1a5200e137bbb2c8691f75a9230cb90528f9b07dabR46) a `runs-on` or `runner`, because the builds can all happen on a CPU runner (`linux.c7i.2xlarge`). This way we don't use the expensive GPU runners for just building monarch. Is there a difference when monarch is built on a GPU runner vs a CPU runner?

.github/workflows/ci.yml
jithunnair-amd: > What's up with these tags? I thought your summary mentioned "labels" on PRs instead?

There's a little bit of "magic" that happens with the labels converting into tags. This is a well-established technique we use in the `pytorch/pytorch` repo as well. Essentially, any PR that gets a `ciflow/rocm` _label_ will trigger the [pytorchci bot](https://github.com/pytorch/test-infra/blob/e950957c2aef5ac196b373ad29bf1e15692e2a21/torchci/lib/bot/ciflowPushTrigger.ts#L15) to create a `ciflow/rocm/*` _tag_ (where the '*' is the PR number). You can see tags matching this pattern in: https://github.com/meta-pytorch/monarch/tags .... This was set up by the PyTorch CI infra team, and we have been successfully using it in [pytorch/pytorch](https://github.com/pytorch/pytorch/blob/e67851696160592edeff3fc166218a707a712b3c/.github/workflows/trunk.yml#L10) and other projects to trigger ROCm CI only when certain labels are added. This helps to avoid triggering CI on _every_ update on _every_ PR. The tags are automatically cleaned up once the PR is merged. Ccing @huydhn from PyTorch CI infra team to add/correct my explanation of how this works.

.github/workflows/ci.yml
jithunnair-amd: That's actually our intention too, which is why the triggers for `ci.yml` are:
```
  pull_request:
    branches:
      - main
      - gh/**
  push:
    branches:
      - main
```

However, this comment is (probably confusingly) trying to explain that we prevent the CPU jobs from running _twice_ on the same PR:
* as a result of the `pull_request` trigger (which _any_ PR update will do)
* as a result of the `ciflow/rocm/` tag push trigger (which updates to PRs with `ciflow/rocm` label will do)

You can see this in action on this PR's runs itself, since this PR has the `ciflow/rocm` label:
* [`pull_request` trigger](https://github.com/meta-pytorch/monarch/actions/runs/23768937534): runs CPU jobs
* [`ciflow/rocm/` tag push trigger](https://github.com/meta-pytorch/monarch/actions/runs/23768939452): skips CPU jobs

For a regular developer, all the jobs that are run as part of either trigger above will show up in the unified list of jobs at the end of the PR page e.g.:
Image

jithunnair-amd: @dulinriley Would it be better if we reworded the comment to:

"CPU jobs skip on ciflow/rocm tag push trigger since they already run for the pull_request trigger for all PRs"

.github/workflows/ci.yml
jithunnair-amd: Since this artifact name is specific to `cuda`, and we are triggering a test matrix in `test-gpu-python.yml`, we moved the artifact naming to https://github.com/meta-pytorch/monarch/pull/3190/changes#diff-9e21efbb5f47d8ee4206cc13f2bb92fbdeeb439bb51b24927544a22bc2580d27R27, where we can use the `${{ matrix.name }}` to give each artifact a unique name.

.github/workflows/set-matrix.yaml
jithunnair-amd: Actually @zstreet87, I was going to suggest renaming `matrix` to `test-matrix`, which is used for the test jobs, and to clearly distinguish it from `build-matrix`, which is used for build jobs. @dulinriley, if you would rather include the python version aspect in the matrix name, we could call it `all-python-versions-build-matrix`, but it might be too wordy.

jithunnair-amd: Also, we're [not building with ROCm for all python versions](https://github.com/meta-pytorch/monarch/pull/3190/changes#diff-5e89c9711a76fb25b73cde1a5200e137bbb2c8691f75a9230cb90528f9b07dabR49), so it might be a bit misleading to call it `all-python-versions-build-matrix` unless we do so for ROCm too. Maybe it's the comment below that needs tweaking to be factually correct.

.github/workflows/set-matrix.yaml
jithunnair-amd: We run CUDA jobs too for _every_ PR, but as part of the `pull_request` trigger (please read [this ](https://github.com/meta-pytorch/monarch/actions/runs/23768937534) for a detailed explanation of the `pull_request` vs `ciflow/rocm` tag triggers). This is seen in https://github.com/meta-pytorch/monarch/actions/runs/23768937534 run for this PR. We can reword the comment to explicitly refer to the triggers instead of "PR run" and "ROCm tag" if you think that would be clearer.

.github/workflows/set-matrix.yaml
jithunnair-amd: In that case, we will want to update https://github.com/meta-pytorch/monarch/pull/3190/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR76 to exclude ROCm jobs triggered on PRs (via the `ciflow/rocm` tags) from blocking the merge. cc @zstreet87

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,84 @@
...
+          ROCM_71='{"name": "rocm7.1", "runner": "${{ inputs.runner-rocm }}", "gpu-arch-type": "rocm", "gpu-arch-version": "7.1", "docker-image": "pytorch/almalinux-builder:rocm7.1", "torch-spec": "torch --extra-index-url https://download.pytorch.org/whl/nightly/rocm7.1/"}'
+          ROCM_72='{"name": "rocm7.2", "runner": "${{ inputs.runner-rocm }}", "gpu-arch-type": "rocm", "gpu-arch-version": "7.2", "docker-image": "pytorch/almalinux-builder:rocm7.2", "torch-spec": "torch --extra-index-url https://download.pytorch.org/whl/nightly/rocm7.2/"}'
+
+          # Define build matrices (includes python-version)
+          CUDA_PY310='{"name": "cuda", "python-version": "3.10", "docker-image": "pytorch/manylinux2_28-builder:cuda12.8", "torch-spec": "--pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu128"}'
+          CUDA_PY312='{"name": "cuda", "python-version": "3.12", "docker-image": "pytorch/manylinux2_28-builder:cuda12.8", "torch-spec": "--pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu128"}'
```
jithunnair-amd, line 48: @zstreet87 Why do the CUDA entries use `manylinux2_28-builder` docker images, but ROCm uses `almalinux-builder` images?

.github/workflows/build-gpu.yml
dulinriley: I haven't tested that the built artifact is identical. I suppose the only thing that matters is if the same cuda/rocm libraries can be installed on a machine without a GPU.
If you can show that there is no difference in the built artifact that would be good and can save some GPU capacity

.github/workflows/ci.yml
dulinriley: Maybe our stance is a bit different from the pytorch main ecosystem. I think I would prefer if we support ROCm to run it on every commit we can. Excluding it from PRs by default should already save quite a bit of capacity (since every commit needs at least one PR update). It's hard to know ahead of time whether a commit might accidentally break the ROCm build

.github/workflows/ci.yml
dulinriley: I see, the tag push happens on the PR branch? Perhaps the tag-based workflow could be an "AND" of both a PR + has tag, instead of either?

.github/workflows/build-gpu.yml
jithunnair-amd: Should be easy by comparing the sizes and/or content of the built artifact for a CPU vs GPU runner. Let me work on getting that out.

.github/workflows/set-matrix.yaml
jithunnair-amd: Updated to use manylinux2_28 images for ROCm

.github/workflows/build-rocm.yml
zstreet87: yeah please see: https://github.com/meta-pytorch/monarch/pull/2891

.github/workflows/ci.yml
jithunnair-amd: > I think I would prefer if we support ROCm to run it on every commit we can.

https://github.com/meta-pytorch/monarch/pull/3190/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR8-R10 ensures exactly that.
```
  push:
    branches:
      - main
```

So ROCm CI build *and* tests will run on *every* commit to `main`, as well as on PRs that have the `ciflow/rocm` label.

.github/workflows/ci.yml
jithunnair-amd: Yes, the tags are pushed on the PR branches. However, every matching trigger will invoke a new workflow run; that's just how Github Actions does it, so there's no way for us to "combine" the two workflow runs. Does that explain it better?

.github/workflows/set-matrix.yaml
jithunnair-amd: @dulinriley Can you please confirm if you want [`status-check`](https://github.com/meta-pytorch/monarch/blob/3b0cc781f5f152ce43fc1d847208165c6e3905c6/.github/workflows/ci.yml#L70) to exclude ROCm CI jobs?

.github/workflows/test-gpu-python.yml
jithunnair-amd: Yes, we do, but that's handled by [`set-matrix.yaml`](https://github.com/meta-pytorch/monarch/pull/3190/changes#diff-5e89c9711a76fb25b73cde1a5200e137bbb2c8691f75a9230cb90528f9b07dabR43)

jithunnair-amd:
@dulinriley While we wait for the ROCm test jobs on this PR to successfully run (to get a green signal), can you please address/resolve some of the outstanding comments?

jithunnair-amd:
After a few more fixes, we got successful results for the ROCm test jobs as part of the `ci.yml` triggered by the `ciflow/rocm` tag push trigger: https://github.com/meta-pytorch/monarch/actions/runs/23886461865/job/69652323992?pr=3190!

dulinriley:
We just released v0.4.1, so we're now clear to add stuff to CI. I'll take another look at this shortly

dulinriley:
Also heads up that I have an open change to add CUDA 13.0 to our test matrix:
https://github.com/meta-pytorch/monarch/pull/3339
That PR does some work to generalize the build commands for multiple versions. It's not a high priority so I won't merge it until other stuff goes in.
I can merge that after this one since yours came first, and I'll try to incorporate the ROCm version into that.

.github/workflows/build-gpu.yml
```diff
@@ -0,0 +1,37 @@
+name: Build GPU
+
+on:
+  workflow_call:
```
dulinriley, line 4: I think this job should be parameterized. Take as input which GPU arch + version + python version to test.
Then the matrix should be set higher (in ci.yml) and then calls this job with that matrix, since I believe the "strategy" can be applied at any level in Github Actions?

.github/workflows/build-gpu.yml
dulinriley: If you can confirm the build outputs are identical that would be great, we can save some GPU machine usage.

.github/workflows/ci.yml
```diff
@@ -73,15 +77,7 @@ jobs:
...
-             [[ "${{ needs.test-gpu-python.result }}" != "success" ]] ||
-             [[ "${{ needs.test-cpu-rust.result }}" != "success" ]] ||
-             [[ "${{ needs.test-cpu-rust-arm64.result }}" != "success" ]] ||
-             [[ "${{ needs.test-gpu-rust.result }}" != "success" ]]; then
-            echo "One or more jobs failed"
-            exit 1
-          else
-            echo "All jobs passed"
-          fi
+        if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
```
dulinriley, line 84: Are there any other statuses than "failure", "cancelled", and "success"? If there are, is there a way to rewrite this as `all(needs.*.result, 'success')` instead?

.github/workflows/set-matrix.yaml
dulinriley: I see, we're distinguishing the set of options applied at build time and test time. In theory either could be more general or more restricted depending on goals.
Ok just make sure there's a comment explaining why each feature is in the build vs. test matrix

.github/workflows/set-matrix.yaml
dulinriley: Yeah sorry the tag "ciflow/rocm" vs PR has got me confused here. It would probably be best if we could describe these as "OR" conditions so that we don't worry about accidentally scheduling a job twice. Or, let it happen and have "concurrency" cancel it?

.github/workflows/set-matrix.yaml
dulinriley: Yes for now status-check should not block a PR from merging if the ROCm CI fails. We can change that once we have confidence in the stability of the CI, which just needs maybe a month of historical data off main (let's say < 1% failure rate when other jobs on the same commit were green?).

We're very sensitive to the speed of merging and we want these jobs to be reliable. Having to check whether our PR broke this workflow is time consuming.

.github/workflows/test-gpu-python.yml
```diff
@@ -2,36 +2,29 @@ name: Test GPU Python
 
 on:
   workflow_call:
-    inputs:
-      artifact-name:
```
dulinriley, line 6: I recommend keeping artifact-name, and pass it in via a matrix from above. That way it keeps very explicit what dependencies test-gpu-python.yml has on the last job.

dulinriley (approved):

dulinriley:
Once CI passes on this we can import and merge

.github/workflows/set-matrix.yaml
jithunnair-amd: Yes, so we now have `build-matrix` and `test-matrix` to disambiguate what each one is for. Any other info such as python versions would stay in the definition of the matrices, to allow us the flexibility of increasing/decreasing coverage as desired.

zstreet87:
Hey @dulinriley, I'm thinking the failures I'm seeing in my PR are also breaking in main. Can you take a look? My PR should be good to go

dulinriley:
yes I'd say that those tests are not failing due to ROCm or any build issue, you should be fine to go

.github/workflows/ci.yml
jithunnair-amd: From what I can clean from search results, if the statuses are only one of "success", "failure", or "cancelled", this would be equivalent, but if any jobs "skipped", that would not be counted as success either and fail the status check.

PR #3329, opened 2026-03-31
Back out "Fix bugs in doc_build deploy" and "Add version switcher to docs with multi-version deployment"

dulinriley:
Summary:
Version switcher worked locally but broke when deployed to Github Pages via `gh-pages`
branch.
I'll leave the branch around since it already has correctly built artifacts, but we need to get the
website back.

Differential Revision: D98992849

### DIFF
--- a/.github/workflows/doc_build.yml
+++ b/.github/workflows/doc_build.yml
@@ -3,19 +3,10 @@ name: Docs
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


dulinriley:
I was able to fix the website on gh-pages luckily, so I don't need to revert. Site is working now with a version picker!

PR #3344, opened 2026-04-01
Fix missing fields in MacOS build

mkuchnik:
MacOS has additional file fields that cause the build to fail:

```
      error[E0063]: missing fields `crtime` and `flags` in initializer of `fuse3::path::reply::FileAttr`
         --> monarch_extension/src/readonly_fuse.rs:375:24
          |
      375 |             .unwrap_or(FileAttr {
          |                        ^^^^^^^^ missing `crtime` and `flags`

      error[E0063]: missing fields `crtime` and `flags` in initializer of `fuse3::path::reply::FileAttr`
         --> monarch_extension/src/readonly_fuse.rs:161:8
          |
      161 |     Ok(FileAttr {
          |        ^^^^^^^^ missing `crtime` and `flags`

      error[E0063]: missing fields `crtime` and `flags` in initializer of `fuse3::path::reply::FileAttr`
         --> monarch_extension/src/chunked_fuse.rs:173:8
          |
      173 |     Ok(FileAttr {
          |        ^^^^^^^^ missing `crtime` and `flags`

      For more information about this error, try `rustc --explain E0063`.
```

on v0.4.0 this is:
```
 error[E0063]: missing fields `crtime` and `flags` in initializer of
      `fuse3::path::reply::FileAttr`
        --> monarch_extension/src/chunked_fuse.rs:83:8
         |
      83 |     Ok(FileAttr {
         |        ^^^^^^^^ missing `crtime` and `flags`

      For more information about this error, try `rustc --explain E0063`.
```

The first commit was tested on top of v0.4.0 (and thus can be cherry-picked into the tagged v0.4.0). The second was necessary on top of main. All changes are guarded on MacOS builds so no behavior should change for other platforms.

### DIFF
--- a/monarch_extension/src/readonly_fuse.rs
+++ b/monarch_extension/src/readonly_fuse.rs
@@ -171,6 +171,10 @@ fn extract_stat_attr(dict: &Bound<'_, PyDict>) -> PyResult<FileAttr> {
         gid: get(dict, "st_gid")?,
         rdev: 0,
         blksize: 4096,
+        #[cfg(target_os = "macos")]
+        crtime: UNIX_EPOCH,
+        #[cfg(target_os = "macos")]
+        flags: 0,
     })
 }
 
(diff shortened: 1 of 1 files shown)


dulinriley:
Thank you for this PR! Sorry I didn't see it earlier. I don't think we need to patch into 0.4.0 because we don't have a MacOS binary release, so people need to build from source anyways

mkuchnik:
> Thank you for this PR! Sorry I didn't see it earlier. I don't think we need to patch into 0.4.0 because we don't have a MacOS binary release, so people need to build from source anyways

Yeah Linux wheels should be unaffected but the 0.4.0 tag in the source code may not build on MacOS.

PR #3357, opened 2026-04-03
snapshot introspection: BFS capture with typed fold seam and visited-set dedup (#3357)

shayne-fletcher:
Summary:

this diff adds the capture layer in monarch_introspection_snapshot. it introduces capture.rs, defines SnapshotData as the typed per-snapshot accumulator, and adds capture_snapshot, which performs a bfs from NodeRef::Root, resolves each typed node reference at most once, projects each NodePayload through the existing convert_node path, and folds the result through a single SnapshotData::push_converted seam. the traversal stays imperative, while subtype routing is confined to the fold target.

the new module also documents and tests the capture invariants directly. the tests use stub resolvers rather than a live mesh and cover the intended semantics: one snapshot row per capture, capture-start timestamp distinct from node as_of, typed visited-set dedup, per-parent edge emission order, duplicate nodes producing one node row but multiple child edges, full NodeKindRow routing through the fold seam, resolver failures aborting capture, and successfully resolved NodeProperties::Error payloads producing ResolutionErrorRow rather than failing capture. lib.rs, BUCK, and the shared to_micros visibility were updated accordingly to wire the new module in cleanly.

Reviewed By: allenwang28

Differential Revision: D99473110

### DIFF
--- a/monarch_introspection_snapshot/Cargo.toml
+++ b/monarch_introspection_snapshot/Cargo.toml
@@ -15,3 +15,4 @@ monarch_record_batch = { version = "0.0.0", path = "../monarch_record_batch" }
 
 [dev-dependencies]
 hyperactor = { version = "0.0.0", path = "../hyperactor" }
+tokio = { version = "1.50.0", features = ["full", "test-util", "tracing"] }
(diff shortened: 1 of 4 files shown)


allenwang28 (approved):

PR #3358, opened 2026-04-03
expose DatabaseScanner::ingest_batch as public ingestion API

shayne-fletcher:
Summary:
this diff adds the narrow public ingestion seam needed by the snapshot bridge without changing telemetry ingestion semantics. it exposes DatabaseScanner::ingest_batch as a thin public wrapper over the existing internal named-batch insertion path, keeping ingestion ownership in monarch_distributed_telemetry while avoiding direct construction or manipulation of LiveTableData from the bridge crate.

it also adds focused unit tests that pin down the contract of that public seam: first ingest creates a table from the batch schema, empty batches still register schema without appending rows, repeated ingests to the same table reuse the existing entry and accumulate rows, and the “empty first, non-empty later” path works correctly. that empty-batch schema-registration behavior is the key guarantee the snapshot ingestion path needs so all snapshot tables can be registered in DataFusion even when a given capture has zero rows for some table.

Differential Revision: D99477096

### DIFF
--- a/monarch_distributed_telemetry/src/database_scanner.rs
+++ b/monarch_distributed_telemetry/src/database_scanner.rs
@@ -313,10 +313,9 @@ impl DatabaseScanner {
 }
 
 impl DatabaseScanner {
-    /// Static method to push a batch to the table_data map.
-    /// This can be used from closures that capture the Arc.
-    ///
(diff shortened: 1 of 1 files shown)


allenwang28 (approved):

PR #3364, opened 2026-04-05
remotemount: fix FUSE refresh race and eliminate full chunk copy

zdevito:
* __->__ #3364

Previously, refresh() copied the entire chunk buffer (up to 11GB) on every
sync even when only a few blocks changed, causing hot refreshes to take 9-19s.
A separate but critical correctness bug: metadata and chunk data were protected
by two independent locks, so a FUSE read could observe new chunk data with old
file offsets/sizes (or vice versa).

Rust (chunked_fuse.rs):
- Merge FsMetadata (RwLock) + ChunkStorage (Mutex<Vec<u8>>) into a single
  FsState struct under one Arc<RwLock<FsState>> containing both the metadata
  HashMap and the flat chunk buffer.
- All FUSE reads hold the single read lock across both the metadata lookup
  and the data copy — no window between the two operations.
- refresh() acquires the write lock once, applies dirty block patches to
  state.chunks, then swaps state.metadata — both atomically. No reader can
  see a partial update.
- refresh() now takes chunk_buf (full mmap memoryview), dirty_ranges
  (list of (offset, length) pairs), new_total_size, and chunk_size instead
  of copying the whole buffer unconditionally.
- Remove update_chunk_range() and resize_chunks() from PyMountHandle (chunk
  updates now happen inside refresh() under the write lock).

Python (remotemount.py):
- FUSEActor tracks _pending_dirty_blocks (list[int] | None) across the
  transfer phase. None signals a full buffer copy (initial mount).
- receive_blocks, replace_blocks, receive_block: accumulate dirty block
  indices into _pending_dirty_blocks instead of calling update_chunk_range.
- _do_refresh: converts _pending_dirty_blocks to (offset, length) pairs and
  passes them to refresh() for atomic application with the metadata swap.
- mount endpoint: sets _pending_dirty_blocks = None so _do_refresh does a
  full buffer copy (transfers ran before _fuse_handle existed).

Result: hot refresh fuse_refresh drops from ~8.78s to ~0.76s (dominated by
the 600ms TTL sleep for kernel cache expiry). Race condition eliminated.

Also:
- fast_pack.py: add directory walk and pack_files_with_offsets timing,
  switch fragmentation/layout log messages to stderr prints.
- mount_config.py: log elapsed time for mount/refresh operations.
- gather_mount.py: use prepare_mount_point instead of os.makedirs to handle
  stale FUSE mounts.
- cli.py: delete cached state file after monarch kill to prevent stale
  cache errors on the next run.

Differential Revision: [D99616713](https://our.internmc.facebook.com/intern/diff/D99616713)

### DIFF
--- a/python/tests/test_remotemount.py
+++ b/python/tests/test_remotemount.py
@@ -1373,11 +1373,12 @@ class TestFuseRefresh:
                     f.write(b"updated content!")
                 meta2, staging_mv2, chunks2, _h2, _pi2 = pack_directory_chunked(src)
                 chunk_size2 = len(bytes(chunks2[0])) if chunks2 else 1
+                buf2 = staging_mv2 if staging_mv2 is not None else memoryview(b"")
                 handle.refresh(
                     meta2,
-                    staging_mv2,
-                    [(0, len(staging_mv2))],
-                    len(staging_mv2),
+                    buf2,
+                    [(0, len(buf2))],
+                    len(buf2),
                     chunk_size2,
                 )
 


zdevito:
@zdevito has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D99616713).

PR #3369, opened 2026-04-06
make TableStore::ingest_batch async to avoid leaking block_in_place into downstream Rust callers (#3369)

shayne-fletcher:
Summary:

made TableStore::ingest_batch async so the new Rust-facing ingestion path no longer inherits the internal block_in_place bridge from DatabaseScanner.

the underlying table mutation is already async via LiveTableData::push. keeping TableStore::ingest_batch synchronous meant async Rust callers were forced through a hidden runtime constraint and needed a multi-thread tokio runtime. this change keeps the existing sync telemetry path intact and only makes the TableStore seam honest for downstream Rust callers.

updated the TableStore tests to await ingestion directly. downstream snapshot ingestion can now await ingest_batch normally instead of relying on the blocking bridge.

Reviewed By: allenwang28

Differential Revision: D99680546

### DIFF
--- a/monarch_distributed_telemetry/src/database_scanner.rs
+++ b/monarch_distributed_telemetry/src/database_scanner.rs
@@ -139,13 +139,37 @@ pub struct TableStore {
 }
 
 impl TableStore {
+    /// Create an empty standalone table store.
+    ///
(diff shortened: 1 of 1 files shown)


allenwang28 (approved):

PR #3372, opened 2026-04-06
[monarch] Enable tensor engine with CPU

dulinriley:
* __->__ #3372

Differential Revision: [D99519774](https://our.internmc.facebook.com/intern/diff/D99519774/)

Enable the tensor engine on CPU-only builds (including macOS) by splitting the monolithic `tensor_engine` feature into two layers. `tensor_engine` now covers the worker actor, the worker message protocol, and tensor plumbing, and it builds on any platform. `tensor_engine_gpu` adds NCCL, RDMA, rdmaxcel, and the static C++ libraries, and it is Linux+GPU only.

The message protocol previously required `torch-sys-cuda` because `ReduceOp` and `UniqueId` lived there. We move those types into a new `monarch_types::nccl_types` module as wire-compatible stubs, so CPU workers can deserialize NCCL messages that arrive from GPU peers. The CUDA-aware constructors (for example, `UniqueId::new_nccl`) live behind a `UniqueIdExt` trait gated on the `cuda` feature of `torch-sys-cuda`.

`nccl-sys` becomes optional on `torch-sys-cuda`, with new `nccl_common` and `nccl_stubs` modules providing the non-CUDA path. We propagate `default-features = false` and `cuda` feature gates through `monarch_messages`, `monarch_tensor_worker`, `monarch_rdma`, and `monarch_extension`, and we adjust the matching BUCK targets to gate CUDA-only deps behind `select({"DEFAULT": [...], "ovr_config//os:macos": []})`.

Runtime adjustments keep things working when there is no GPU. `WorkerActor` probes `torch.cuda.is_available()` before creating a `CudaDevice`. NCCL-initiating methods (`BackendNetworkInit`, `SplitComm`, `send_tensor`, and friends) short-circuit when no device or communicator exists. The Python client (`client.py`) skips `BackendNetworkInit`, `SplitComm`, and `BackendNetworkPointToPointInit` when CUDA is unavailable. `rdma.py` stubs the native bindings with an `_UnavailableNativeBinding` placeholder and reports `rdma_supported() == False` instead of raising at import. `monarch.common._C` (`init.cpp`) guards `patch_cuda`, `mock_cuda`, and `unmock_cuda` behind `MONARCH_HAS_CUDA`. `tensor.check_cuda` no-ops without CUDA. A new `_rust_bindings.has_tensor_engine()` export replaces the RDMA-import probe in `_check_tensor_engine_available`.

macOS CI and the nightly wheel build now run with the tensor engine enabled. `build-cpu-macos.yml`, `test-cpu-python-macos.yml`, `test-cpu-rust-macos.yml`, and the nightly `build-macos.yml` either set `USE_TENSOR_ENGINE=1` or drop the override entirely.

Rust test modules that depend on `caffe2:torch` (`monarch_tensor_worker::borrow`, `comm`, `lib`, `stream`, and `monarch_types::pytree`) are gated behind `cfg(all(test, fbcode_build))` so they are excluded from OSS builds. `test_tensor_engine.py` gains `needs_tensor_engine` and `needs_cuda` markers; `test_actor_error.py` skips two cases off Linux.

Test Plan:
- macOS CI (`build-cpu-macos`, `test-cpu-python-macos`, `test-cpu-rust-macos`) all pass with `USE_TENSOR_ENGINE=1`.
- Nightly `build-macos` wheel builds and `python -c "import monarch"` succeeds.
- Linux CI (`build-cuda`, `build-cpu`, `test-cpu-python`, `test-cpu-rust`, `test-gpu-python`, `test-gpu-rust`) remains green.
- `uv run pytest python/tests/test_tensor_engine.py` on macOS: `needs_cuda` cases skip, the remaining cases exercise the CPU tensor engine.

### DIFF
--- a/python/monarch/_src/rdma/rdma.py
+++ b/python/monarch/_src/rdma/rdma.py
@@ -11,13 +11,18 @@ import logging
 import sys
 import warnings
 from collections import defaultdict
-from typing import Any, cast, List, Optional, Tuple, TYPE_CHECKING
+from enum import Enum
+from typing import Any, cast, Dict, List, Optional, Tuple, TYPE_CHECKING
 
 if TYPE_CHECKING:
     import torch
 
 from monarch._rust_bindings.monarch_hyperactor.pytokio import PythonTask, Shared
-from monarch._src.actor.proc_mesh import ProcMesh
+from monarch._src.actor.actor_mesh import Actor, context
+from monarch._src.actor.endpoint import endpoint
+from monarch._src.actor.future import Future
+from monarch._src.actor.proc_mesh import get_or_spawn_controller, ProcMesh
+from pyre_extensions import none_throws
 from typing_extensions import Self
 
 _NATIVE_RDMA_IMPORT_ERROR: Optional[ImportError] = None
@@ -49,16 +54,6 @@ except ImportError as e:
 
     def _rdma_supported() -> bool:
         return False
-
-
-from enum import Enum
-from typing import Dict
-
-from monarch._src.actor.actor_mesh import Actor, context
-from monarch._src.actor.endpoint import endpoint
-from monarch._src.actor.future import Future
-from monarch._src.actor.proc_mesh import get_or_spawn_controller
-from pyre_extensions import none_throws
 
 
 # RDMARead/WriteTransferWarnings are warnings that are only printed once per process.


dulinriley:
ghstack has gotten pretty confused with this so I will close and re-open

PR #3382, opened 2026-04-08
snapshot service: add spawn_periodic_capture with overlap guard and deterministic PT-* tests

shayne-fletcher:
Summary:
adds periodic snapshot capture to monarch_introspection_snapshot via spawn_periodic_capture, with a factored private loop, per-tick overlap control through the shared in_flight CAS guard, cancellation via CancellationToken, and error logging without terminating the loop.

tightens the service module documentation to explain one-shot versus periodic capture and the resolver shapes they require, and adds PT-* coverage for zero-interval rejection, missing live sink rejection, delayed first fire, overlap skipping, cancellation boundaries, failure resilience, and live-ingest-only behavior. the test suite is now deterministic for the loop-driven cases through explicit notify-based synchronization, with one narrow real-timer test retained to cover the actual interval_at(now + interval, interval) behavior.

Differential Revision: D100024224


allenwang28 (approved):

Issue #1215, opened 2025-09-15
Complete GPU-GPU RDMA support.


johnwhumphreys:
Assign to Dennis V once he's on the OSS project.

dstaay-fb:
Introduce GPU<>(GPU/CPU) support in Python APIs on RDMA.   Included integration into PyTorch's CudaCachingAllocator.


Related PRs: 
https://github.com/meta-pytorch/monarch/pull/1200
https://github.com/meta-pytorch/monarch/pull/1224

Guido1Alessandro1Trevisan:
@dstaay-fb Shouldn’t there topology awareness logic if two GPUs are in the same node? RIght now data between two gpus in the same node is transferred via InfiniBand when you use RDMA. Wouldn’t it be better to transfer data through NVLink using NVIDIA GPU P2P instead of InfiniBand?

dstaay-fb:
> [@dstaay-fb](https://github.com/dstaay-fb) Shouldn’t there topology awareness logic if two GPUs are in the same node? RIght now data between two gpus in the same node is transferred via InfiniBand when you use RDMA. Wouldn’t it be better to transfer data through NVLink using NVIDIA GPU P2P instead of InfiniBand?

You're intuition matches mine, just initial use cases targeted inter-node comms.

dstaay-fb:
for transparency initial Python APIs suggested by @zdevito

```
LocalMemory = torch.Tensor | memoryview

class RDMAAction:
    """
    Schedule a bunch of actions at once. This corresponds to the functionality in ibv_post_send,
    but generalized over qp.
    """
    def read(dst: LocalMemory | List[LocalMemory], src: RDMABuffer) -> Self:
        """
            *dst = src
        If dst is a list, it is the concatenation of the data in the list
        """
        ...
    
    def write(dst: RDMABuffer, src: LocalMemory | List[LocalMemory]) -> Self:
        """
            *dst = src
        
        If src is a list, it is the concatenation of the data in the list.
        """
        ...
    
    def fetch_add(dst: RDMABuffer, orig: LocalMemory, add: int) -> Self:
        """
        Atomically:
            *orig = *dst
            *dst = *dst + add
        orig/dest are 8 bytes
        """
        ...
    
    def compare_and_swap(dst: RDMABuffer, orig: LocalMemory, compare: int, swap: int)
        """
        Atomically:
            *orig = *dst;
            if (*dst == compare) {
                *dst = swap    
            }
[...]

PR #3391, opened 2026-04-10
EFA RDMA: P4d compat, loopback fix, GPU Direct, polling optimization

### LINKED ISSUE #3376: RDMABuffer.read_into() fails between actors on same node (EKS/EFA)

dmvevents:
## Summary

Five EFA fixes for Monarch's ibverbs RDMA backend, plus an engineering guide.

### Fix 1: Same-node loopback memcpy fallback (fixes #3376)
EFA SRD silently drops same-node loopback RDMA traffic. Added memcpy fallback in `IbvBackend::execute_op()` for CPU buffers.

### Fix 2: WriteWithImm warning on EFA
EFA emulates `fi_writedata` as send/recv — receiver pool exhausts after ~240 ops. Added `tracing::warn`.

### Fix 3: GPU Direct RDMA validation for EFA (relates to #1215)
`validate_execution_context()` checked for Mellanox-only `nvidia_peermem` and `PeerMappingOverride`. Now checks `efa_nv_peermem` on EFA devices.

### Fix 4: Adaptive CQ polling (performance)
Replaced fixed 1ms sleep with adaptive spin→yield→sleep backoff. The 1ms floor added unnecessary latency for transfers completing in microseconds on EFA.

### Fix 5: Selective signaling for multi-chunk transfers (performance)
Only signal the last chunk in `put()`/`get()` instead of every chunk. Reduces CQ entries from N to 1 for >1 GiB transfers.

### Docs: EFA RDMA engineering guide
500-line guide from 221 CUCo instincts covering SRD protocol, loopback, MR lifetime, CQ polling, GPU Direct, proxy architecture.

## Changes
- `manager_actor.rs` — memcpy fallback + adaptive polling
- `queue_pair.rs` — WriteWithImm warning + selective signaling
- `rdma_components.rs` — EFA-aware `validate_execution_context()`
- `ibv_manager_actor_tests.rs` — loopback tests
- `docs/EFA_RDMA_LEARNINGS.md` — engineering guide

## Test plan
- [ ] `test_rdma_{write_from,read_into}_local_same_actor_loopback`
- [ ] Existing RDMA tests pass
- [ ] On EFA: `validate_execution_context()` checks `efa_nv_peermem`
- [ ] Adaptive polling: small transfers complete without 1ms floor
- [ ] Selective signaling: multi-chunk transfers produce 1 CQ entry

### DIFF
--- a/docs/EFA_RDMA_LEARNINGS.md
+++ b/docs/EFA_RDMA_LEARNINGS.md
@@ -1,0 +1,500 @@
+# EFA RDMA Engineering Guide for Monarch
+
+> Distilled from 198 instincts (89 EFA/RDMA-specific, 52 hardware-verified) accumulated
+> during the CUCo optimization campaign on AWS P5.48xlarge (H100, 4x EFA) and
+> P4d.24xlarge (A100, 4x EFA). Each section below links to the source instinct IDs
+> in `knowledge_base/learned_instincts/instincts.json`.
+
+---
+
+## 1. SRD Is Not InfiniBand — What Breaks
+
+EFA uses the Scalable Reliable Datagram (SRD) protocol. Three InfiniBand assumptions
+fail silently on SRD:
+
+### 1.1 No message ordering
+
+SRD delivers RDMA writes out of order. Posting writes A, B, C may deliver C, A, B.
+Any protocol that chains operations assuming order-of-completion will silently corrupt
+data or deadlock.
+
+**Monarch impact**: `IbvBackend::execute_op` processes ops sequentially within a batch,
+which is safe. But if future work pipelines overlapping writes to adjacent memory regions,
+the lack of ordering becomes a correctness hazard.
+
+### 1.2 CQ completion is local-only
+
+A send CQ completion means the local NIC accepted the WQE into its send queue. It does
+**not** mean the remote side received the data. `ibv_post_send` can even return success
+for packets that SRD will silently drop (e.g., loopback).
+
+> *I203 [confidence 10, ground-truth]*: `ibv_post_send` returns 0 even when SRD will
+> drop the packet. The return code only confirms the NIC accepted the WQE.
+
+**Monarch impact**: The `wait_for_completion` loop polls the send CQ, which is correct
+for confirming the local side is done. But callers must not treat this as proof that the
+remote buffer was written.
+
+### 1.3 No RDMA atomics
+
+EFA does not support `IBV_ACCESS_REMOTE_ATOMIC`. Monarch's `efa::mr_access_flags()`
+correctly omits this. The `RDMAAction.fetch_add` and `compare_and_swap` methods will
+fail on EFA without a software fallback.
+
+> *I047 [confidence 11, ground-truth]*: All 7 attempts to replace barriers with
+> alternative synchronization (fence, quiet, int_p, epoch) failed on EFA.
+
+### 1.4 FI_MORE is broken
+
+The libfabric `FI_MORE` hint (batch operations before triggering the NIC) does not work
+on the `efa-direct` provider. Every attempt to use it — counter-based flush, progress-based
+flush, timer-based flush — caused hangs. 15+ attempts, zero successes.
+
+> *I010 [confidence 3, ground-truth]*: FI_MORE does NOT work on EFA efa-direct provider.
+> All patches hang.
+
+### 1.5 fi_writedata is emulated and exhaustible
+
+EFA RDM emulates `fi_writedata` (RDMA write with immediate) as a send/recv pair.
+After ~240 writes, the receiver buffer pool exhausts and operations stall. Use
+`fi_write` or `fi_writemsg` instead.
+
+> *I128 [confidence 11, ground-truth]*: fi_writedata receiver buffer pool exhausts
+> after ~240 writes.
+
+**Monarch impact**: The `WriteWithImm` operation in `post_op_efa` maps to `fi_writedata`
+under the hood (via the EFA provider). The existing warning in `post_op_efa` for
+`WriteWithImm` is correct — it should not be used in hot loops on EFA.
+
+---
+
+## 2. Same-Node Loopback
+
+**The #1 production gotcha on EFA.**
+
+SRD silently drops packets destined for the same node. The full QP lifecycle completes
+without error (INIT → RTR → RTS), `ibv_post_send` returns success, but the CQ completion
+never arrives.
+
+### 2.1 How it manifests
+
+```
+error: broken link: failed to enqueue in MailboxClient
+  channel closed with reason Some("delivery timeout")
+```
+
+Or from the Rust layer:
+
+```
+[buffer] rdma operation did not complete in time (expected wr_ids=[0])
+```
+
+Everything *looks* connected. The QP is RTS. The post returns 0. But `poll_cq` returns
+0 forever because SRD dropped the packet at the NIC.
+
+### 2.2 Current fix (PR #3391)
+
+`IbvBackend::execute_op` detects same-actor loopback and uses `memcpy`:
+
+```rust
+if crate::efa::is_efa_device()
+    && self.actor_id() == op.remote_manager.actor_id()
+{
+    // CPU buffers: direct memcpy
+    std::ptr::copy_nonoverlapping(src, dst, size);
+}
+```
+
+For CUDA device memory, the check falls through with a warning (needs `cudaMemcpy`).
+
+### 2.3 Remaining gap: cross-process same-node
+
+Two actors in **different** processes on the same node have different `IbvManagerActor`
+instances. The actor-ID check misses this case. Known solutions:
+
+1. Compare GIDs at QP connection time (same GID = same node)
+2. Shared-memory handshake between node-local processes
+3. Fall back to TCP when EFA loopback is detected at connect time
+
+> *I204 [confidence 9, ground-truth]*: Cross-process same-node EFA loopback is unsolved
+> in Monarch as of PR #3391.
+
+> *I176 [confidence 10, ground-truth]*: Memcpy fallback pattern for EFA same-process
+> loopback.
+
+---
+
+## 3. Memory Registration
+
+### 3.1 CPU tensors: allocator can pull the rug
+
+PyTorch's caching allocator may reclaim and reallocate the underlying memory between
+`ibv_reg_mr` and the RDMA operation. This causes `IBV_WC_LOC_LEN_ERR` (status=1,
+vendor_err=104) on large tensors (≥10 MB) under memory pressure.
+
+The MR was registered pointing at virtual address X with size S. By the time the RDMA
+write executes, the allocator freed and reused that memory. The NIC tries to DMA from
+an address that no longer holds the expected data — or worse, the MR itself was
+invalidated.
+
+**Mitigations**:
+
+- Hold a strong reference to the tensor for the entire RDMA operation lifetime
+- Use `torch.Tensor.pin_memory()` for buffers involved in RDMA
+- Re-register the MR immediately before the operation if the tensor might have moved
+- Enable `expandable_segments` so the caching allocator uses stable large blocks
+
+> *I202 [confidence 7, hypothesis]*: Monarch #1136 large tensor flakiness likely stems
+> from MR/allocator mismatch under memory pressure.
+
+### 3.2 GPU memory: DMA-BUF constraints
+
+DMA-BUF registration works on EFA P5 instances (kernel 6.17+). Monarch already implements
+this path in `register_mr` via `ibv_reg_dmabuf_mr`. Constraints discovered through testing:
+
+| Constraint | Detail |
+|-----------|--------|
+| Separate `cudaMalloc` per NIC | Sub-ranges of one allocation → CQ err=22 (EINVAL) |
+| `nvshmem_malloc` incompatible | DMA-BUF requires `cudaMalloc` memory specifically |
+| `efa_nv_peermem` module required | Must be loaded; check `/proc/modules` |
+| `FI_EFA_USE_DEVICE_RDMA=1` | Required env var for GPU Direct RDMA |
+| libfabric ≥ 1.18 | For FI_HMEM_CUDA support |
+
+> *I080 [confidence 3, hypothesis]*: DMA-BUF for multi-NIC GPU RDMA requires SEPARATE
+> cudaMalloc per NIC.
+>
+> *I122 [confidence 10, ground-truth]*: cudaMalloc staging buffers bypass the
+> nvshmem_malloc MR issue.
+>
+> *I076 [confidence 3, hypothesis]*: DMA-BUF GPU memory registration works on EFA P5
+> kernel 6.17.
+
+### 3.3 GPU L2 cache coherence after RDMA
+
+GPU L2 cache is **not** coherent with incoming RDMA writes. After remote data arrives
+via RDMA, the GPU must call `cuFlushGPUDirectRDMAWrites` before reading it. Without the
+flush, the GPU may read stale L2-cached data from before the RDMA transfer.
+
+> *I129 [confidence 11, ground-truth]*: cuFlushGPUDirectRDMAWrites required after RDMA
+> data arrives, before GPU reads.
+
+**Monarch impact**: If Monarch adds GPU tensor RDMA support, the completion path must
+insert a GPU-side flush before returning the buffer to the caller.
+
+---
+
+## 4. Completion Queue Polling
+
+### 4.1 Never skip CQ polls
+
+On EFA (via libfabric), `fi_cq_read` is not just a status check — it drives the
+provider's internal progress engine. Skipping polls when you think nothing is pending
+prevents the provider from completing other in-flight operations.
+
+Monarch's `wait_for_completion` (poll every 1ms) is correct. Do not "optimize" by
+reducing poll frequency or skipping empty polls.
+
+> *I020 [confidence 8, ground-truth]*: Adaptive CQ polling breaks dispatch by starving
+> the progress engine.
+
+### 4.2 CQ iteration sweet spot
+
+For NCCL GIN on EFA, `CQ_ITER=64` is a verified local optimum. The curve is U-shaped:
+
+| CQ_ITER | BW (GB/s) | Delta |
+|---------|-----------|-------|
+| 32 | 2.048 | -2.7% |
+| **64** | **2.308** | **baseline** |
+| 128 | 1.906 | -17.4% |
+
+> *I186 [confidence 8, ground-truth]*: CQ_ITER=128 regresses LL by -17.4%. 64 is the
+> sweet spot.
+
+### 4.3 Adding CQ-only workers does not help
+
+The EFA bottleneck is `fi_write` issuance (the actual RDMA post), not CQ polling. Adding
+dedicated CQ polling threads gives 0% throughput improvement.
+
+> *I057 [confidence 10, ground-truth]*: CQ-only proxy workers give 0% throughput gain.
+
+### 4.4 fi_cntr is thread-local on EFA
+
+EFA's hardware completion counter (`fi_cntr`) is thread-local. Calling `fi_write` from
+thread A and `fi_cntr_read` from thread B returns invisible completions. If Monarch ever
+uses multi-threaded RDMA submission, completions must be polled from the **same thread**
+that posted the operation.
+
+> *I056 [confidence 11, ground-truth]*: EFA fi_cntr is thread-local.
+
+---
+
+## 5. GPU Direct RDMA on EFA
+
+### 5.1 It works — and it's fast
+
+FI_HMEM_CUDA (GPU Direct RDMA via libfabric) is functional on EFA P5 instances.
+Standalone benchmarks show EFA **matches or exceeds** InfiniBand:
+
+| Scenario | EFA | InfiniBand | Ratio |
+|----------|-----|------------|-------|
+| 4-NIC standalone, 64 MB writes | 44 GB/s | 43 GB/s | 1.02x |
+| Dispatch simulation (4096×7168 BF16) | 45.1 GB/s | 43 GB/s | 1.05x |
+| Single NIC, 256 KB writes | 8.2 GB/s | — | — |
+
+The hardware is not the bottleneck. The gap between EFA and InfiniBand in real workloads
+comes from the CPU proxy architecture, not the NIC itself.
+
+> *I116 [confidence 11, ground-truth]*: FI_HMEM_CUDA WORKS on P5 EFA.
+>
+> *I083 [confidence 3, hypothesis]*: 45.1 GB/s dispatch simulation exceeds IB reference.
+
+### 5.2 Requirements checklist
+
+For Monarch to enable GPU Direct RDMA on EFA:
+
+```
+[ ] libfabric >= 1.18 installed (current EFA installer includes 2.4.0+)
+[ ] efa_nv_peermem kernel module loaded (check: grep efa_nv_peermem /proc/modules)
+[ ] FI_EFA_USE_DEVICE_RDMA=1 in environment
+[ ] Kernel >= 6.17 for DMA-BUF fd support
+[ ] GPU memory allocated via cudaMalloc (not nvshmem_malloc or managed memory)
+[ ] Separate cudaMalloc per NIC for multi-NIC configs
+[ ] cuFlushGPUDirectRDMAWrites after RDMA receive, before GPU read
+```
+
+### 5.3 The peermem symbol version issue (Monarch #2672)
+
+`efa_nv_peermem` can fail to load after NVIDIA driver updates because DKMS rebuilds the
+NVIDIA modules but not `efa_nv_peermem`. Fix: rebuild `efa_nv_peermem` after any NVIDIA
+driver change, or pin both driver and peermem to compatible versions in your container image.
+
+> *I071 [confidence 11, ground-truth]*: FI_HMEM_CUDA registration requires
+> efa_nv_peermem module loaded.
+
+---
+
+## 6. The Proxy Architecture Gap
+
+### 6.1 Why EFA needs a CPU proxy
+
+InfiniBand supports IBGDA (InfiniBand GPU Direct Async) — the GPU posts RDMA operations
+directly to the NIC via PCIe MMIO writes to hardware doorbells. EFA does not support IBGDA.
+
+On EFA, a CPU thread must mediate every RDMA operation:
+1. Read data source address from GPU (or CPU) memory
+2. Post `ibv_post_send` / `fi_writemsg` to the EFA NIC
+3. Poll CQ for completion
+4. Signal the requestor that the operation finished
+
+This CPU round-trip is where the EFA-vs-InfiniBand latency gap lives. The NIC itself is
+comparably fast (45 GB/s EFA vs 43 GB/s IB at the hardware level).
+
+### 6.2 EFA uses PROXY_MINIMAL
+
+On EFA, NVSHMEM's proxy mode is `PROXY_MINIMAL`. There are no proxy channels. All code
+that optimizes multi-proxy channel distribution is dead code on EFA. Enhanced proxy
+(adding `progress_channels` to the minimal loop) has zero effect.
+
+> *I118 [confidence 11, ground-truth]*: EFA uses PROXY_MINIMAL.
+>
+> *I119 [confidence 10, ground-truth]*: Enhanced proxy has zero effect on throughput.
+
+### 6.3 The proven high-performance pattern (pplx)
+
+The fastest known EFA RDMA implementation (pplx-garden, 63 GB/s) uses:
+
+1. **GPU kernel** writes tokens to a registered send buffer
+2. **GPU signals CPU** via GDRCopy `st_mmio_b8` (MMIO write, ~1 μs)
+3. **CPU worker** calls `fi_writemsg` from the registered staging memory
+4. **CPU polls CQ** for completion
+5. **CPU signals GPU** via `nvshmem_signal` that the transfer finished
+
+The key insight: the CPU reads a ring buffer **after** the kernel completes, not during.
+Concurrent GPU-CPU access to the same memory causes hangs because workers cannot access
+GPU memory while CUDA kernels are executing on the same device.
+
+> *I073 [confidence 10, ground-truth]*: pplx achieves 63 GB/s with this post-dispatch
+> architecture.
+>
+> *I112 [confidence 3, hypothesis]*: Workers CANNOT access GPU memory while CUDA kernels
+> are executing on the same device.
+>
+> *I113 [confidence 3, hypothesis]*: Post-dispatch architecture WORKS — kernel writes to
+> device ring during execution, CPU reads ring AFTER cudaStreamSynchronize.
+
+### 6.4 Batched writes dominate throughput
+
+RDMA chunk size is the single most important tuning variable for normal-mode throughput.
+Transfer size matters more than batch count:
+
+| Chunk size | BW (GB/s) | Speedup |
+|-----------|-----------|---------|
+| 4 KB | 3.7 | 1.0x |
+| 116 KB | 31.4 | 8.5x |
+| 256 KB | 40+ | 10.8x |
+
+> *I133 [confidence 10, ground-truth]*: RDMA chunk size is the dominant normal-mode
+> tuning variable (8.5x range).
+>
+> *I130 [confidence 10, ground-truth]*: Batched 256KB writes: 16MB → 40 GB/s,
+> 64MB → 44 GB/s. At 64MB, EFA exceeds InfiniBand.
+
+**Monarch impact**: The current `MAX_RDMA_MSG_SIZE = 1 GiB` with chunking in `put()`/`get()`
+is fine for correctness. But for performance-sensitive paths, experimenting with chunk
+sizes in the 256 KB–16 MB range would likely improve throughput.
+
+---
+
+## 7. EFA Configuration Reference
+
+### 7.1 IbverbsConfig defaults for EFA
+
+Monarch applies these via `efa::apply_efa_defaults()`:
+
+| Parameter | EFA Value | Mellanox Default | Why |
+|-----------|----------|-----------------|-----|
+| `gid_index` | 0 | 3 | EFA uses GID index 0 |
+| `max_send_sge` | 1 | 30 | EFA limit |
+| `max_recv_sge` | 1 | 30 | EFA limit |
+| `max_dest_rd_atomic` | 0 | 16 | No RDMA atomics on EFA |
+| `max_rd_atomic` | 0 | 16 | No RDMA atomics on EFA |
+
+### 7.2 RDMA buffer sizing constraints
+
+| Parameter | Safe Range | Failure Mode |
+|-----------|-----------|--------------|
+| `rdma_chunk_size` | ≤ `buffer_size / 2` | Config assertion crash |
+| `rdma_buffer_size` | 512 optimal (with 1 GB backing) | 1024+ regresses |
+| `num_rdma_bytes` | ≤ 1 GB | 2 GB → OOM on H100 80 GB |
+| `OFI_NCCL_EAGER_MAX_SIZE` | Default only | EFA provider rejects 32768 |
+
+> *I017, I018, I019 [confidence 10, ground-truth]*: Buffer sizing constraints verified
+> on P5 hardware.
+>
+> *I151 [confidence 11, ground-truth]*: EFA rejects eager max 32768.
+
+### 7.3 EFA provider selection
+
+Two providers exist: `efa` and `efa-direct`. On P5:
+
+- `efa` (aws-patches): 10% **better** throughput for normal dispatch
+- `efa-direct` (aws-patches-2): Higher raw bandwidth but worse dispatch latency
+
+> *I054 [confidence 9, ground-truth]*: efa-direct gives 10% WORSE throughput than efa
+> for normal dispatch.
+
+### 7.4 P5.48xlarge NUMA topology
+
+```
+NUMA 0: CPU cores 0-47, 16 EFA NICs (rdmap0s*-rdmap15s*)
+NUMA 1: CPU cores 48-95, 16 EFA NICs (rdmap16s*-rdmap31s*)
+```
+
+Pin RDMA proxy threads to the same NUMA node as the NIC they drive. Cross-NUMA NIC
+access adds measurable latency.
+
+> *I132 [confidence 10, ground-truth]*: P5.48xlarge has 2 NUMA nodes, not 4.
+
+---
+
+## 8. What This Means for Monarch
+
+### 8.1 Current EFA support status
+
+EFA ibverbs support merged in PR #2638 (Feb 2026). Key components:
+
+- **Auto-detection**: `efadv_query_device()` → `is_efa_device()` cached in `OnceLock`
+- **QP connect**: `rdmaxcel_efa_connect()` handles INIT → RTR → RTS + address handle
+- **Post ops**: `rdmaxcel_qp_post_op()` for write/read/recv
+- **Loopback fix**: PR #3391 adds memcpy fallback for same-process loopback
+
+> *I178 [confidence 10, ground-truth]*
+
+### 8.2 CUDA decoupling is already mostly done
+
+Three coupling points exist, all of which gracefully degrade without CUDA:
+
+| Coupling Point | What It Does | Without CUDA |
+|---------------|-------------|-------------|
+| `cuPointerGetAttribute` | Detect CUDA vs CPU memory | Returns "not CUDA" → CPU path used |
+| `pytorch_segment_scanner` | `torch.cuda.memory._snapshot()` | Returns 0 segments → skipped |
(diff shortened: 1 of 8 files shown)


docs/EFA_RDMA_LEARNINGS.md
```diff
@@ -0,0 +1,500 @@
...
+
+A send CQ completion means the local NIC accepted the WQE into its send queue. It does
+**not** mean the remote side received the data. `ibv_post_send` can even return success
+for packets that SRD will silently drop (e.g., loopback).
+
+> *I203 [confidence 10, ground-truth]*: `ibv_post_send` returns 0 even when SRD will
+> drop the packet. The return code only confirms the NIC accepted the WQE.
+
+**Monarch impact**: The `wait_for_completion` loop polls the send CQ, which is correct
+for confirming the local side is done. But callers must not treat this as proof that the
```
samlurye, line 35: Is there anything that could be added that would allow monarch to detect that the remote buffer was written?

docs/EFA_RDMA_LEARNINGS.md
```diff
@@ -0,0 +1,500 @@
...
+> in Monarch as of PR #3391.
+
+> *I176 [confidence 10, ground-truth]*: Memcpy fallback pattern for EFA same-process
+> loopback.
+
+---
+
+## 3. Memory Registration
+
+### 3.1 CPU tensors: allocator can pull the rug
```
samlurye, line 129: I feel a little uncertain about this one. We already explicitly hold a strong reference to the tensor for the lifetime of the rdma op unless the user explicitly calls `RdmaBuffer.drop()`.

From the Nvidia docs on `IBV_WC_LOC_LEN_ERR`: "This event is generated when the receive buffer is smaller than the incoming send. It is generated on the receiver side of the connection."

If the memory had genuinely been freed, it seems more likely to me that this would manifest as data corruption,

docs/EFA_RDMA_LEARNINGS.md
```diff
@@ -0,0 +1,500 @@
+# EFA RDMA Engineering Guide for Monarch
```
samlurye, line 1: It would be super awesome if you could share with us your setup that enabled you to discover all of this!

docs/EFA_RDMA_LEARNINGS.md
```diff
@@ -0,0 +1,500 @@
...
+> *I122 [confidence 10, ground-truth]*: cudaMalloc staging buffers bypass the
+> nvshmem_malloc MR issue.
+>
+> *I076 [confidence 3, hypothesis]*: DMA-BUF GPU memory registration works on EFA P5
+> kernel 6.17.
+
+### 3.3 GPU L2 cache coherence after RDMA
+
+GPU L2 cache is **not** coherent with incoming RDMA writes. After remote data arrives
+via RDMA, the GPU must call `cuFlushGPUDirectRDMAWrites` before reading it. Without the
```
samlurye, line 175: I've been trying to read the docs to get some clarity on this. On https://docs.nvidia.com/cuda/gpudirect-rdma/:

> Only CPU initiated CUDA APIs provide ordering of GPUDirect memory operations as observed by the GPU. That is, despite a third party device having issued all PCIE transactions, a running GPU kernel or copy operation may observe stale data or data that arrives out-of-order until a subsequent CPU initiated CUDA work submission or synchronization API. To ensure that memory updates are visible to CUDA kernels or copies, an implementation should ensure that all writes to the GPU BAR happen before control is returned to the CPU thread which will invoke the dependent CUDA API.

And from the documentation for [cuFlushGPUDirectRDMAWrites](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__DEVICE.html#group__CUDA__DEVICE_1g265e3c82ef0f0fe035f85c4c45a8fbdf):

> If the scope equals or lies within the scope indicated by [CU_DEVICE_ATTRIBUTE_GPU_DIRECT_RDMA_WRITES_ORDERING](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gge12b8a782bebe21b1ac0091bf9f4e2a3dfb3df2c0e9347a7c13f71043e961c50), the call will be a no-op and can be safely omitted for performance. This can be determined by comparing the numerical values between the two enums, with smaller scopes having smaller values.

The first link makes it sound like as long as the remote write has completed w.r.t the CPU on the destination host, then any subsequent CUDA work launched from the CPU will see the result of the write. But the second link makes me wonder if this is always sufficient -- if we look at the documentation for [CUGPUDirectRDMAWritesOrdering](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g8d647023012ffc318ce86814fec05376):

> CU_GPU_DIRECT_RDMA_WRITES_ORDERING_NONE = 0
The device does not natively support ordering of remote writes. [cuFlushGPUDirectRDMAWrites()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__DEVICE.html#group__CUDA__DEVICE_1g265e3c82ef0f0fe035f85c4c45a8fbdf) can be leveraged if supported.

Meanwhile, the torchcomms ibverbs implementation doesn't have any references to `cuFlushGPUDirectRDMAWrites`. They appear to rely on a [local RDMA read to force a PCI-e flush](https://fburl.com/code/sgfjc7on).

Do you have any insight into when `cuFlushGPUDirectRDMAWrites` is truly necessary?

docs/EFA_RDMA_LEARNINGS.md
```diff
@@ -0,0 +1,500 @@
...
+The EFA bottleneck is `fi_write` issuance (the actual RDMA post), not CQ polling. Adding
+dedicated CQ polling threads gives 0% throughput improvement.
+
+> *I057 [confidence 10, ground-truth]*: CQ-only proxy workers give 0% throughput gain.
+
+### 4.4 fi_cntr is thread-local on EFA
+
+EFA's hardware completion counter (`fi_cntr`) is thread-local. Calling `fi_write` from
+thread A and `fi_cntr_read` from thread B returns invisible completions. If Monarch ever
+uses multi-threaded RDMA submission, completions must be polled from the **same thread**
```
samlurye, line 224: This is a good callout -- every time the CQ poll loop hits an `await`, it could resume execution on a different thread.

docs/EFA_RDMA_LEARNINGS.md
```diff
@@ -0,0 +1,500 @@
...
+### 4.1 Never skip CQ polls
+
+On EFA (via libfabric), `fi_cq_read` is not just a status check — it drives the
+provider's internal progress engine. Skipping polls when you think nothing is pending
+prevents the provider from completing other in-flight operations.
+
+Monarch's `wait_for_completion` (poll every 1ms) is correct. Do not "optimize" by
+reducing poll frequency or skipping empty polls.
+
+> *I020 [confidence 8, ground-truth]*: Adaptive CQ polling breaks dispatch by starving
```
samlurye, line 197: Not sure I understand this one -- the adaptive polling never skips polls?

monarch_rdma/src/backend/ibverbs/manager_actor.rs
```diff
@@ -1100,6 +1115,19 @@ impl IbvBackend {
...
+    /// Returns `true` if `addr` points to CUDA device memory.
+    fn is_device_memory(addr: usize) -> bool {
+        unsafe {
+            let mut mem_type: i32 = 0;
+            let err = rdmaxcel_sys::rdmaxcel_cuPointerGetAttribute(
+                &mut mem_type as *mut _ as *mut std::ffi::c_void,
+                rdmaxcel_sys::CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
+                addr as rdmaxcel_sys::CUdeviceptr,
+            );
+            err == rdmaxcel_sys::CUDA_SUCCESS
```
samlurye, line 1127: I don't think this check is sufficient to determine if it's device memory or not -- you would also need to check `mem_type == CU_MEMORYTYPE_DEVICE`. But there is actually already a helper function that does this: `monarch_rdma::local_memory::is_device_ptr`.

monarch_rdma/src/backend/ibverbs/manager_actor.rs
```diff
@@ -1108,6 +1136,56 @@ impl IbvBackend {
...
+                            std::ptr::copy_nonoverlapping(
+                                remote_addr as *const u8,
+                                local_addr as *mut u8,
+                                size,
+                            );
+                        }
+                    }
+                }
+                return Ok(());
+            }
```
samlurye, line 1186: Can we factor this block out into its own helper function?

monarch_rdma/src/backend/ibverbs/queue_pair.rs
```diff
@@ -533,18 +533,26 @@ impl IbvQueuePair {
...
         while remaining > 0 {
             let chunk_size = std::cmp::min(remaining, MAX_RDMA_MSG_SIZE);
+            let is_last_chunk = chunk_size >= remaining;
             let idx = unsafe {
                 rdmaxcel_sys::rdmaxcel_qp_fetch_add_send_wqe_idx(
                     self.qp as *mut rdmaxcel_sys::rdmaxcel_qp,
                 )
             };
-            wr_ids.push(idx);
+            // Only signal the last chunk to reduce CQ pressure. For single-chunk
```
samlurye, line 542: How much does this actually help? It makes me a little nervous -- if an early chunk fails, then my (potentially wrong) understanding is that all we'll see on the last chunk is `IBV_WC_WR_FLUSH_ERR` with no way to access the real root cause.

monarch_rdma/src/efa.rs
```diff
@@ -56,6 +59,53 @@ fn is_efa_device_impl() -> bool {
...
+/// When this returns `false` on an EFA device, the ibverbs backend cannot
+/// be used for data transfer and should fall back to TCP transport.
+pub fn efa_supports_rdma() -> bool {
+    *EFA_RDMA_CAPABLE_CACHE.get_or_init(efa_supports_rdma_impl)
+}
+
+fn efa_supports_rdma_impl() -> bool {
+    if !is_efa_device() {
+        // Not EFA — RDMA capability is determined by the device type (RC/UD)
+        return true;
```
samlurye, line 76: Should this case return false?

rdmaxcel-sys/src/rdmaxcel.c
```diff
@@ -669,6 +669,30 @@ int rdmaxcel_is_efa_dev(struct ibv_context* ctx) {
...
+// P4d EFA only supports send/recv — RDMA read/write will fail.
+// Returns 1 if RDMA read/write is supported, 0 if not.
+int rdmaxcel_efa_supports_rdma(struct ibv_context* ctx) {
+  if (!ctx || !ctx->device) {
+    return 0;
+  }
+
+  // Query device capabilities to check for RDMA read/write support.
+  // The reliable way is to check ibv_query_device_ex for max_rdma read/write
+  // Work Requests. If max_sq_rdma > 0, the device supports RDMA operations.
```
samlurye, line 683: Do we need to check `max_sq_rdma` in this function?

docs/EFA_RDMA_LEARNINGS.md
zdevito: One scenario to check:
```
local_tensor = ...
event = remote_buffer.read_from(local_tensor)
del local_tensor
await event
```
Did local_tensor stay alive through the remote read?

docs/EFA_RDMA_LEARNINGS.md
zdevito: Since we haven't defined CUDA stream semantics for our GPU RDMA operations yet, I don't think we are the ones responsible for calling cuFlushGPUDirectRDMAWrites in any case. Once we define a way of synchornizing the RDMABuffer event with a cuda stream, then we have to figure out if we need to run the flush as part of the synchronization or if a local flush read would do the flush.

docs/EFA_RDMA_LEARNINGS.md
samlurye: Would using `FI_DELIVERY_COMPLETE` on writes gives us this behavior?

docs/EFA_RDMA_LEARNINGS.md
samlurye: <validated in a separate thread>

dmvevents:
Thanks for the thorough review @samlurye @zdevito. Working through the asks now:

**`FI_DELIVERY_COMPLETE` on writes (EFA_RDMA_LEARNINGS.md:35)** — No, this won't work on EFA. SRD's CQ completion is local-only (NIC accepted the WQE, not that the remote received the data). `FI_DELIVERY_COMPLETE` is not in SRD's supported transfer-completion modes. The only way to confirm remote visibility on SRD is an application-level ack or a fi_atomic MAX signal used as an ordering marker. Will update the doc to make this explicit.

**`is_device_memory` check (manager_actor.rs:1127)** — Agreed, checking `CUDA_SUCCESS` alone isn't sufficient; will switch to `monarch_rdma::local_memory::is_device_ptr` which does the `mem_type == CU_MEMORYTYPE_DEVICE` check you pointed to.

**Factor out the memcpy block (manager_actor.rs:1186)** — Will extract into a helper.

**Selective signaling (queue_pair.rs:542)** — Fair concern. The perf win is meaningful on ≥1 GiB transfers (N CQ entries → 1) but you're right that an early-chunk failure will only surface as `IBV_WC_WR_FLUSH_ERR` on the last chunk with no root cause. Two options I'm considering:
- Signal every Kth chunk + last (bounds the blast radius of unsignaled failures)
- Keep selective signaling but emit all unsignaled WR IDs into a ring buffer so a flush error can be correlated back to the failing WR
Which would you prefer? I can implement either.

**`max_sq_rdma` check (rdmaxcel.c:683)** — Yes, we should guard against exceeding SQ capacity. The current posting path doesn't bound-check when posting multiple WRs in a loop. Will add the guard.

**`efa.rs:76` fallback return** — Should return `false` in that path; will fix.

**Adaptive polling vs. "never skip polls" (EFA_RDMA_LEARNINGS.md:197)** — Good catch on the ambiguity. The adaptive polling doesn't skip calls to `poll_cq`; it only changes the sleep duration between polls. The doc warning was about skipping polls entirely (which starves the progress engine). I'll reword the doc section so this distinction is clear.

**CUDA flush responsibility (EFA_RDMA_LEARNINGS.md:175)** — @zdevito agreed — until we've defined CUDA stream semantics for `RDMABuffer` events, this isn't our call to make. Will reframe that section as a future concern tied to the stream-sync work, not something this PR needs to solve.

**Tensor lifetime across await (EFA_RDMA_LEARNINGS.md:129)** — You noted this was validated in a separate thread; I'll remove that section from the doc since it no longer applies.

**Setup for discovery (EFA_RDMA_LEARNINGS.md:1)** — Happy to write that up. Our setup was 2× P4d.24xlarge on EKS with EFA, `fi_pingpong` as the starting harness, then incremental `fi_write`/`fi_writemsg` experiments with `FI_LOG_LEVEL=debug`. I'll add a short "How we arrived at this" section with the exact `fi_info`/env settings.

Next push will include: `is_device_ptr` swap, `efa.rs:76` fallback, `max_sq_rdma` guard, memcpy helper extraction, doc rewording. Holding selective-signaling change pending your preference between the two options above.

dmvevents:
Pushed [`a922565`](https://github.com/meta-pytorch/monarch/commit/a922565) with the items I could act on without waiting on your signaling preference:

- **`is_device_memory` → `local_memory::is_device_ptr`** (manager_actor.rs:1127) — dropped the local helper with the incomplete check.
- **Memcpy block extracted** (manager_actor.rs:1186) — now `try_loopback_memcpy(&op)`; the RDMA-path `execute_op` reads cleanly and the memcpy branch is isolated for future testing.
- **`efa.rs:76` fallback** — `efa_supports_rdma_impl` now returns `false` on non-EFA hardware. The sole caller (`rdma_components.rs:123`) already gates on `is_efa_device()`, so no behavior change on that path; the function's contract is now unambiguous.
- **`rdmaxcel.c:683` RDMA capability check** — reworked to consult `efadv_device_attr.device_caps` with `EFADV_DEVICE_ATTR_CAPS_RDMA_READ | _RDMA_WRITE`. The `max_sq_rdma` field you asked about isn't a direct member of `efadv_device_attr` in the header we build against (closest is `max_sq_wr` which is total SQ depth, not RDMA-specific), but `device_caps` is the authoritative signal the EFA driver actually sets for P5/P5en vs P4d. Left the `max_qp_rd_atom > 0` gate as a defensive check in case IBV disagrees with the EFA driver.
- **Doc rewordings (EFA_RDMA_LEARNINGS.md):**
  - 3.1 rewritten — you're right, `IBV_WC_LOC_LEN_ERR` is fundamentally a receive-buffer-sizing error per the NVIDIA docs, and `RdmaBuffer` already holds a strong tensor reference through the op. Removed the allocator "pull the rug" framing.
  - 3.3 (GPU L2 flush) — reframed as future work gated on defining CUDA stream semantics for `RDMABuffer` events. Not this PR's problem, per @zdevito.
  - 4.1 (adaptive polling) — clarified that adaptive polling never skips polls; only the inter-poll sleep duration is tuned. The "never skip polls" warning now correctly targets stopping `fi_cq_read` entirely (the libfabric-specific failure mode), not the tunable sleep.

Still open, awaiting your call:

- **Selective signaling (queue_pair.rs:542)** — two options from my earlier reply: (A) signal every K-th chunk + last, or (B) keep selective signaling but emit unsignaled WR IDs into a ring buffer so flush errors are correlatable. Happy to implement either.
- **"How we arrived at this" write-up (EFA_RDMA_LEARNINGS.md:1)** — will add in the next push.

Let me know on the signaling preference and I'll roll it into the next push with the setup section.

dmvevents:
A few corrections and follow-ups on my earlier reply — an internal multi-model audit caught two things I got wrong:

### 1. Retracting the `fi_atomic MAX` claim

In my earlier response I wrote:

> "The only way to confirm remote visibility on SRD is an application-level ack or a fi_atomic MAX signal used as an ordering marker."

The second half is wrong on two counts:

- **EFA SRD does not expose `FI_ATOMIC`** — the EFA provider's `fi_info` caps exclude atomics entirely. There is no `fi_atomic` on this hardware.
- **SRD has no message ordering**, so even if an atomic were available, it could overtake the payload it was meant to order.

The correct answer to @samlurye's original question ("can monarch detect that the remote buffer was written?"): **on SRD, the only reliable mechanism is a receiver-originated application-level ack.** The receiver must observe the written data (or a sentinel at its tail) and then send an explicit ack back. Nothing sender-side — no completion mode, no atomic, no flag — gives remote-visibility guarantees on this provider.

Apologies for the misstatement.

### 2. Comment [5] — cross-thread `await` resumption (`manager_actor.rs:230`)

I didn't address this one directly. The concern: `wait_for_completion` yields via `tokio::task::yield_now().await` and `tokio::time::sleep(...).await`, and tokio may resume the future on a different executor thread. If the CQ-poll state migrates, is that safe?

Short answer: **yes, by design.** `IbvQueuePair` intentionally stores the verbs pointers as `usize` (see `monarch_rdma/src/backend/ibverbs/queue_pair.rs:100-110`) rather than `NonNull<ibv_qp>` or `&mut`, precisely so the struct is `Send`, `Sync`, `Clone`, and `Serialize`. The safety contract is documented in the struct's header comment:

> The `qp` field stores a pointer to `rdmaxcel_qp_t` (not `ibv_qp`). `rdmaxcel_qp_t` contains atomic counters and completion caches internally. This makes `IbvQueuePair` trivially `Clone` and `Serialize`. Multiple clones share the same underlying `rdmaxcel_qp_t` via the pointer.

Concurrent `poll_cq` calls on the same CQ from different threads are serialized at the `rdmaxcel_qp_t` layer (atomic completion cache), and `ibv_poll_cq` itself is thread-safe per the verbs spec. So a future that resumes on a different tokio worker continues polling the same QP/CQ via the same pointers without invalidating any Rust borrow invariants.

The one thing this design does *not* give you is ordering between concurrent posters on the same QP — but we don't post from multiple threads on a single QP in this PR; posting is gated at the actor boundary, and only polling is concurrent.

### 3. Comment [12] — source tensor lifetime through `await event`

Confirming for the record: the `local_tensor` argument to `read_from(local_tensor)` is held by `RdmaBuffer` via an owned reference until the completion event fires. `del local_tensor` in Python only drops the caller's handle; `RdmaBuffer`'s reference keeps the buffer alive until the RDMA op is acknowledged, so the `await event` scenario is safe. If this invariant is ever relaxed (e.g., `RdmaBuffer.drop()` called early), the guarantee goes away — but under normal usage the tensor stays alive through the remote read.

### 4. Selective signaling — switching my recommendation to Option B

Reviewing this with the full context: **Option B (keep selective signaling + unsignaled-WR ring buffer for diagnosability)** is the better choice, and I'll go with that unless you'd rather not.

Rationale: Option A (every-Kth signaling) tunes a trade-off that isn't obviously useful — you still bound the blast radius by `K` WRs, and it's one more knob to get wrong. Option B preserves the full perf win (1 CQE per ≥1 GiB transfer) while making `IBV_WC_WR_FLUSH_ERR` diagnosable by dumping the last N unsignaled WR IDs into the error message. Zero perf cost on the happy path.

I already drafted both as side branches so you can see them side-by-side:
- **Option A**: [`draft-3391-sig-option-a`](https://github.com/dmvevents/monarch/compare/fix-efa-loopback-3376...draft-3391-sig-option-a) (+24/-7)
- **Option B**: [`draft-3391-sig-option-b`](https://github.com/dmvevents/monarch/compare/fix-efa-loopback-3376...draft-3391-sig-option-b) (+206/-4)

Let me know if B is the call and I'll squash it into the main branch of the PR.

dmvevents:
Status after re-checking against current `main` (2026-08-12): #3376 is still open, and the same-node cross-actor EFA case is still unhandled — main's new `is_loopback` path (manager_actor.rs:473) only covers same-actor + same-device self-QP loopback, not two actors on one node, where EFA SRD silently drops traffic. The backend has since been restructured (#3814, #4039, #4162: `QueuePairInitializer` removed, `IbvDeviceImpl` device abstraction, dedicated `efa_device/efa_domain/efa_queue_pair`), so rather than force-resolving this 3-month-old branch across that refactor, I'll reauthor the surviving fixes (loopback fallback, GDR `efa_nv_peermem` validation, adaptive polling) against the new abstraction, folding in the review feedback already addressed here. If a maintainer would rather I rebase this branch in place, say the word and I'll do that instead.

docs/EFA_RDMA_LEARNINGS.md
dmvevents: Yes on both counts, and they are two different mechanisms worth separating.

`FI_DELIVERY_COMPLETE` is the right answer **on the libfabric path** — it changes what the local completion *means* (delivered to the remote NIC/memory rather than just accepted locally), so `fi_cq_read` becomes proof of remote write. Cost is latency: the completion now waits for a remote ack instead of returning at the local NIC. It is a per-op `fi_writemsg` flag, so it can be opt-in for ops that need the guarantee. Note it still says *delivered to memory*, not *visible to a GPU kernel* — on GPU buffers you additionally need the flush discussed in the GPUDirect thread.

This PR is on the **ibverbs** path, where there is no `FI_DELIVERY_COMPLETE`. The equivalent is `IBV_WR_RDMA_WRITE_WITH_IMM`: the receiver gets a real recv-CQ entry carrying the immediate value, which is a positive remote-side signal rather than an inference. I checked the provider rather than assuming — rdma-core's EFA provider gates `IBV_QP_EX_WITH_RDMA_WRITE_WITH_IMM` on the `RDMA_WRITE` device cap (`providers/efa/verbs.c:1984-1988`) and binds `wr_rdma_write_imm` when requested (`:3264`), so it is available on SRD wherever plain RDMA write is, i.e. P5/P5en but not P4d. It costs a posted recv on the target and a recv CQ, which is a real design change, not a flag flip.

Cheapest option that needs no new capability: a read-back of a sentinel the receiver polls — correct but adds a round trip. Happy to fold the WRITE_WITH_IMM option into the doc as the recommended ibverbs answer if you want it recorded here.

docs/EFA_RDMA_LEARNINGS.md
dmvevents: Agreed, and it is worth stating the concrete consequence in this codebase, because it is sharper than the doc's general warning. `wait_for_completion` awaits at `manager_actor.rs:1062-1068` (`yield_now`, then `sleep`), so under a multi-threaded tokio runtime the loop can and will resume on a different worker thread between polls.

Two things follow. First, the `fi_cntr` thread-locality note (doc section 4.4) means a future libfabric-backed submission path cannot poll a counter across that await — it must keep post and poll on one thread, which in tokio terms means a dedicated blocking thread or a `LocalSet`, not a plain async task. Second, on the current ibverbs path the thing that actually crosses the await is `completion_cache` (`poll_cq_with_cache`, `rdmaxcel.c:348`), which is unsynchronized mutable state reached via a raw pointer; today it is safe only because a single task per QP touches it. That invariant is real but implicit. If you want, I can add it as an explicit note in the doc — 'per-QP poll must stay single-task; the completion cache is not thread-safe' — since it is the kind of thing that gets broken by a later refactor that looks harmless.

docs/EFA_RDMA_LEARNINGS.md
dmvevents: You were right, and the text you were reading was wrong — I have already rewritten it. The original claimed adaptive polling itself was the hazard, which conflated two different things.

Current wording (`docs/EFA_RDMA_LEARNINGS.md:259-264`) says it explicitly: adaptive polling in this PR does **not** skip polls. Every iteration still calls `ibv_poll_cq`; only the sleep between polls is tuned (`manager_actor.rs:1060-1069` — spin ~100 iterations, then 100us, then 1ms). The I020 instinct is about *stopping* `fi_cq_read` and starving libfabric's progress engine, which is a libfabric-path concern; on the ibverbs path we post straight to the NIC and there is no provider progress engine to starve. Section 4.1 is retitled 'Never stop polling the CQ' for that reason. Thanks for pushing on it — the original phrasing would have discouraged a correct optimization.

monarch_rdma/src/backend/ibverbs/manager_actor.rs
dmvevents: You are right on both points and I have taken the helper. `local_memory::is_device_ptr` (`local_memory.rs:29`) already does the full check — `err == CUDA_SUCCESS && mem_type == CU_MEMORYTYPE_DEVICE` — and the loopback path now calls it for both addresses (`manager_actor.rs:1130-1131`); my `is_device_memory` duplicate is gone. Worth being precise about why the original was wrong rather than merely redundant: it declared success on the *API return code* while never reading the `mem_type` it had just written, so host memory registered with the CUDA driver (`cuMemHostRegister`, `CU_MEMORYTYPE_HOST`) would have been classified as device memory and pushed onto the RDMA path instead of the memcpy fallback — the exact case the fallback exists for.

One leftover I should flag rather than let you find it: the same not-comparing-`mem_type` pattern is still in the device-selection code at `manager_actor.rs:448-455`, where `is_cuda` is set from the return code alone. That one is more benign — it only gates a CUDA-ordinal-to-ibv-device lookup that fails closed to the default device — but it is the same latent bug and should use `is_device_ptr` too. Say the word and I will fold that in.

monarch_rdma/src/backend/ibverbs/manager_actor.rs
dmvevents: Done — it is `try_loopback_memcpy` now (`manager_actor.rs:1125`), called from the guard at `:1187`. It returns `bool` with the contract in the doc comment (`true` = loopback satisfied, `false` = fall through to the RDMA path), which also made the CUDA case honest: it warns and returns `false` rather than pretending to have handled it. The extracted version uses `is_device_ptr` per your other comment, and the `copy_nonoverlapping` calls carry the SAFETY note about host-buffer lifetime.

monarch_rdma/src/backend/ibverbs/queue_pair.rs
dmvevents: Your instinct is right about `IBV_WC_WR_FLUSH_ERR`, and it is worth me being straight about the cost/benefit here because I think the honest answer is 'not much'.

On the root-cause concern: it is handled, but only partly and only by accident of a later change. `wait_for_completion` detects `WR_FLUSH_ERR` and re-polls the individual wr_ids looking for a non-flush status to report as the root cause (`manager_actor.rs:1077-1087`). That recovers the real error **when the failing WQE's completion is still in the CQ or the completion cache**. But with last-chunk-only signalling the earlier chunks are unsignalled, so a failure on chunk 3 of 5 may produce no CQ entry we ever look at — and then all we surface is the flush error on the last chunk, exactly as you described. So the drain mitigates the general case and specifically does not fully cover the case this optimization creates.

On the benefit: `MAX_RDMA_MSG_SIZE` is 1 GiB (`queue_pair.rs:16`), so multi-chunk only happens above 1 GiB, and the saving is (n-1) CQ entries on a transfer that is already moving gigabytes. That is not a meaningful CQ-pressure win — the comment I wrote overstates it. I would rather revert to signalling every chunk and keep per-chunk error attribution, which is the more defensible default; if CQ pressure ever shows up in a profile it can come back with a real measurement behind it. Want me to make that change here, or leave it and let it ride as-is?

monarch_rdma/src/efa.rs
dmvevents: Fixed — it returns `false` now (`efa.rs:74-78`), and you caught a genuine inversion. The original returned `true` for non-EFA devices, meaning any RC/UD device would be reported as 'EFA supports RDMA', which is both wrong on its face and wrong in effect: it made a function whose entire purpose is distinguishing P5/P5en from P4d answer optimistically for hardware it knows nothing about.

Returning `false` is the safe direction for a capability probe, with the contract now written down in the docstring (`:67-69`): only meaningful when `is_efa_device()` is true, callers gate on that first, and RC/UD RDMA capability is an independent concern rather than something this function pretends to answer.

rdmaxcel-sys/src/rdmaxcel.c
dmvevents: Good catch, and the answer turned out to be that the field I originally referenced does not exist — the comment named `max_sq_rdma`, but `struct efadv_device_attr` (`infiniband/efadv.h:45-55`) has no such member; the RDMA-related fields are `device_caps` and `max_rdma_size`. So the original was describing a check it was not making.

Current implementation (`rdmaxcel.c:684-705`) uses the capability bits directly, which is the reliable form: `efadv_query_device` then `(device_caps & (EFADV_DEVICE_ATTR_CAPS_RDMA_READ | EFADV_DEVICE_ATTR_CAPS_RDMA_WRITE)) == both`. That matches how rdma-core's own EFA provider gates these ops — it sets `RDMA_WRITE` in `device_caps` (`providers/efa/verbs.c:213`) and only then advertises `IBV_QP_EX_WITH_RDMA_WRITE` on SRD (`:1984-1988`) — so we are reading the same source of truth the provider uses rather than inferring from a queue-depth number. It also checks `ibv_query_device`'s `max_qp_rd_atom != 0` first as a cheap generic gate.

One deliberate strictness note: requiring **both** READ and WRITE means a device with only one is reported unsupported. That is intentional for Monarch's use (`put` and `get` both exist), but if you would rather have per-direction capability I can split it into two accessors.

docs/EFA_RDMA_LEARNINGS.md
dmvevents: @samlurye you were right and I have rewritten the section — it is now 'CPU tensors: MR registration vs buffer sizing' (`docs/EFA_RDMA_LEARNINGS.md:192-206`) and states plainly that `IBV_WC_LOC_LEN_ERR` (status=1, vendor_err=104) fires when the buffer is smaller than the incoming message and is **not** primarily an allocator-lifetime issue, and that Monarch's strong-reference-until-completion behaviour is already sufficient on the lifetime side. Your reasoning was the deciding argument: genuinely freed-and-reused memory would present as corruption, not as a length error, so the original attribution did not fit the evidence.

@zdevito on your scenario — I have **not** run it, and I do not want to imply otherwise. Reading the code, `RdmaBuffer` holds the local memory as `Arc<dyn RdmaLocalMemory>` (`rdma_components.rs:159`) and release is explicit via `drop_buffer`/`release_buffer` (`:219-221`), so the `Arc` should keep `local_tensor` alive across the `del` until the event resolves. But 'should, by inspection' is not the same as verified, and the failure mode if I am wrong is silent corruption rather than an error — exactly the case worth an actual test. The check that would settle it is a Python-level test in that shape (create, `read_from`, `del`, `await`, then assert contents) plus refcount instrumentation to confirm the `Arc` survives the `del`. I would rather add that test than leave the claim resting on my reading; happy to write it if you want it in this PR rather than a follow-up.

docs/EFA_RDMA_LEARNINGS.md
dmvevents: @zdevito that framing is the one I adopted — the section is rewritten to match it (`docs/EFA_RDMA_LEARNINGS.md:240-246`). It now says explicitly that because Monarch has not yet defined CUDA stream semantics for `RDMABuffer` events, this is not yet the right layer to call `cuFlushGPUDirectRDMAWrites`, and that once the synchronization contract exists the choice is between inserting a host flush in the completion path or relying on a local flush-read as a side effect of stream-synchronized reads. It is labelled future work, not something this PR fixes.

@samlurye the doc paragraph you quoted is the crux, and it cuts the way you suspected: the ordering guarantee is tied to *CPU-initiated* CUDA work submission, so a third-party device (the NIC) completing its PCIe writes does not by itself make them visible to a running kernel. That is why the flush cannot live purely on the receiver's GPU side and why it has to be attached to whatever synchronization point we eventually define. Until then the doc tells callers doing GPU RDMA to flush before their first kernel read, so the gap is documented rather than silently inherited.

docs/EFA_RDMA_LEARNINGS.md
dmvevents: Added — there is now a 'How we arrived at this — our debug setup' section at the top of the doc (`docs/EFA_RDMA_LEARNINGS.md:10`, commit `f69584b`). It covers the hardware (2× P5.48xlarge H100/32-EFA on HyperPod EKS, plus a P4d.24xlarge kept specifically as a capability-delta reference — `max_qp_rd_atom = 0` there is what proved 'no atomics' is an EFA property rather than a P4d quirk), the software baseline, and the harnesses in the order we actually used them.

The part most worth stealing is the cheap one: `FI_PROVIDER=efa FI_LOG_LEVEL=debug FI_EFA_USE_DEVICE_RDMA=1 fi_pingpong -e rdm`, run with server and client **on the same host** — that is what first exposed the silent loopback drop, handshake completing with no data completion ever arriving. It needs no application code and reproduces in seconds. Above that we used a small `fi_write`/`fi_writemsg` C harness with a one-hypothesis-per-run rule, which is what kept the negative results (the seven failed barrier-replacement attempts behind I047) attributable instead of confounded. Happy to expand any part of it if something is unclear.

dmvevents:
Status note on the merge conflict, so review effort lands in the right place: `main` has restructured the ibverbs backend since this branch was cut (per-device `efa_device`/`efa_domain`/`efa_queue_pair` + `mlx_*` modules, CQ pooling in #4668/#4688, and #4039 removed the test scaffolding this PR's tests build on). A plain merge is not meaningful anymore — notably, `efa_queue_pair.rs:433-438` on main now asserts `device_caps` RDMA read+write at QP creation, which is exactly the condition this PR converts into a graceful TCP fallback for P4d-class devices.

Plan: I'll re-port the three behaviors (the P4d capability gate ahead of that assert, the same-node loopback memcpy path, and the polling fix) onto the new module layout as fresh commits, and either update this branch or open a successor PR referencing the review threads here. The discussion in the open threads (delivery-complete semantics, `is_wr_flush_err` wiring, `max_rdma_size`) all carries over — none of it is invalidated by the restructure.

dmvevents:
Re-port landed as #4734 (five commits on current main: the P4d gate ahead of the new QP-creation assert, the same-node loopback copy, the `is_wr_flush_err` wiring and `signaled` plumb-through you asked for, and the EFA notes doc). Review context from the threads here carries over — happy to close this one in favor of #4734 once you've had a look, or keep both open if you'd rather compare.

PR #3461, opened 2026-04-17
[hyperactor] replace nested span context with subject-based log prefixes

mariusae:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #3461

Hyperactor log output currently renders nested tracing spans as verbose, hard-to-read prefixes:

```
[spawn_with_name{host_mesh:local, proc_mesh:anon-...}, spawn_inner{actor_id:..., actor_name:..., actor_type:...}, recording{recording:..., recorder:...}] message
```

Even the author of this code has a hard time parsing these. This diff replaces nested span rendering with a clear **subject** prefix that identifies the entity the log pertains to:

```
<actor metatls:host:1234,local,my_actor[0]> message, key:value
<proc local:0,my_proc> event, name:ProcStatus, status:Created
```

Changes:

- Add `Subject` type (`hyperactor/src/subject.rs`) that borrows an ID reference and formats as `<kind id>` (e.g. `<actor ...>`, `<proc ...>`). Extensible for channels/sessions later.
- Add `SUBJECT_KEY` constant in `hyperactor_telemetry` as the well-known tracing field name.
- GlogSink now looks for a `subject` field on events and span chains, rendering it as a prefix instead of dumping the full nested span tree. Other subscribers (Scuba, Perfetto) see it as a regular string field.
- Recording spans accept a `subject` parameter (needed because they use `parent: None`).
- Replace verbose span fields (`actor_id`, `actor_name`, `actor_type`, `host_mesh`, `proc_mesh`) with `subject` on `spawn_inner`, `actor_instance`, `handle_message`, `spawn_with_name`, `allocate_inner`, and `allocate_serve_proc` spans.
- Add `Subject::proc` to proc lifecycle events (`ProcStatus::Created/Dropped`), `destroy_and_wait_except_current`, `handle_unhandled_supervision_event`, and `stop_actor`.

Differential Revision: [D101364390](https://our.internmc.facebook.com/intern/diff/D101364390/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D101364390/)!

PR #3532, opened 2026-04-21
pass test subject to recording.span in test callers

dulinriley:
Summary: Recording::span now requires a &str subject (see D101364390 / PR #3461). Two test callers in monarch_hyperactor were missed by the original fix in D101807697 and still call `recording.span()` with no argument, breaking OSS Rust CI. Pass "test" as the subject.

Reviewed By: thedavekwon

Differential Revision: D101866661


ndmitchell (approved):

PR #3758, opened 2026-05-05
Bump rust-toolchain to 1.95 (#3758)

dulinriley:
Summary:

bumping rust nightly compiler pin to 1.95

Reviewed By: zdevito

Differential Revision: D103763044


dulinriley:
This is being done a different way

PR #3859, opened 2026-05-13
[ROCm] - Fix rocm exclusion function

### LINKED ISSUE #3830: DISABLED python/tests/test_tensor_engine.py::test_proc_mesh_tensor_engine
Error message is:
```
Failed: Error evaluating 'skipif' condition
    7.2.53211
        ^
SyntaxError: invalid syntax
```

alugorey:
Fixes https://github.com/meta-pytorch/monarch/issues/3830

### DIFF
--- a/python/tests/test_tensor_engine.py
+++ b/python/tests/test_tensor_engine.py
@@ -21,7 +21,7 @@ needs_tensor_engine = pytest.mark.skipif(
     reason="tensor_engine was not built",
 )
 needs_cuda = pytest.mark.skipif(
-    not torch.cuda.is_available() or torch.version.hip,
+    not torch.cuda.is_available() or torch.version.hip is not None,
     reason="cross-rank tensor transport requires CUDA+NCCL (not ROCm/HIP)",
 )
 


dulinriley:
Can you explain a bit why that check didn't work before? Was hip set to something that was Falsey but not None?

According to this page:
https://docs.pytorch.org/docs/2.12/notes/hip.html#checking-for-hip

It sounds like we can just only check torch.cuda.is_available() and we don't need to bother checking `hip` at all?

alugorey:
> Can you explain a bit why that check didn't work before? Was hip set to something that was Falsey but not None?
> 
> According to this page: https://docs.pytorch.org/docs/2.12/notes/hip.html#checking-for-hip
> 
> It sounds like we can just only check torch.cuda.is_available() and we don't need to bother checking `hip` at all?

line 24 was expecting to be resolved to a boolean expression. torch.version.hip returned the version number rather than True/False which resulted in a syntax error. I think the link you sent checking for hip works because it's not called from within skipif(). as skipif() is returning the syntax error, seems like it expects the expression to explicitly resolve to a bool whereas the example in your link uses intrinsic pythonisms to check for hip version. Initial motivation seems to have come from here:
https://github.com/meta-pytorch/monarch/commit/b80a3312b4590a0cf6b7d35d15ff39f7f63cfd26

This is the PR that added the check against torch.version.hip. Reading the commit message, looks like the author indeed intended to write `torch.version.hip is not None` but didn't reflect that in the actual code. The reason we are checking against hip version (and not just `torch.cuda.is_available()` is also detailed in the commit message. Seems like `to_mesh()` is a CUDA/NCCL specific feature (though the test passed for me locally on rocm. may have to do a follow up PR as this PR is simply to re-enable the test.) 

tl;dr this change now correctly skips the test on ROCm rather than a syntax error due to torch.version.hip returning a string rather than a bool.

dulinriley (approved):

dulinriley:
Thanks for the context, sorry I didn't check github frequently, just stamped it.
I didn't realize pytest treated a non-bool value as something special

PR #4139, opened 2026-06-02
Disable ROCm CI jobs while ROCm is broken

dulinriley:
Summary: ROCm GPU CI is currently very broken, so disable it in normal CI. `set-matrix.yaml` now emits CUDA-only build and test matrices for PRs, push, and schedule. ROCm still runs only on a manual `ciflow/rocm` tag push, for iterating on fixes before re-enabling.

Reviewed By: thedavekwon

Differential Revision: D107127116


dulinriley:
This disables ROCm CI until https://github.com/meta-pytorch/monarch/issues/4145 is resolved

PR #4281, opened 2026-06-23
SlurmJob: support SLURM batch mode (via `client_script`) (#4281)

JenniferWang:
Summary:
## Summary 
`SlurmJob` previously supported only an *external controller* model: `apply()`
submits an allocation of workers running `run_worker_loop_forever`, and a
separate controller process (a login node, or a second job) calls `state()` to
`attach_to_workers`. Batch mode was explicitly rejected:

```python
def _create(self, client_script):
    if client_script is not None:
        raise RuntimeError("SlurmJob cannot run batch-mode scripts")
```

This PR implements the batch (`client_script`) launch mode that `JobTrait` /
`LocalJob` already model: **one allocation runs the workers *and* the client,
and the allocation is released when the client exits.**

## Motivation
The external-controller model has two downsides:

1. **Placement.** The controller has to live outside the allocation, which is
   awkward to schedule and run.
2. **Dangling workers.** When the controller exits, the worker allocation is
   *not* released. `run_worker_loop_forever` treats a client disconnect as "kill
   current work and wait for a new connection" — it does not exit — and
   `SlurmJob` has no exit-time teardown on this path, so the workers hold their
   nodes until the job's time limit. Only a single allocation gives
   SLURM-enforced teardown, which is what batch mode provides.

## What it does

`apply(client_script=...)`:
- caches a `BatchJob(self)` to `.monarch/job_state.pkl` (same pattern as
  `LocalJob._create`), then
- submits one sbatch whose body is a small in-allocation runner
  (`monarch/_src/job/_slurm_batch.py`).

The runner, on the allocation's head node:
1. seeds `run_worker_loop_forever` on every node (`srun --ntasks-per-node=1`),
2. runs the client as argv (no shell) with `MONARCH_BATCH_JOB=1`,
3. terminates the workers in a `finally`.

The client reconnects via `state()` -> `_load_cached()` -> the cached `BatchJob`
(whose `can_run` is gated on `MONARCH_BATCH_JOB`), so it attaches to the running
workers instead of submitting a new allocation. When the runner exits, the batch
script ends and SLURM frees every node, so the job records **COMPLETED with the
client's exit code** -- no `scancel`/`trap`, no dangling workers.

## Scope / future work

**The controller colocates on the first worker node (it is CPU-only and rides on a
node already allocated for the workers).** 

A dedicated controller node via a SLURM
heterogeneous job is intentionally out of scope for this PR.

## Backward compatibility

The external-controller path (`apply()` with no `client_script`) is unchanged.

Test Plan:
- **Unit** (`python/tests/_src/job/test_slurm.py`, mocking `sbatch`/`squeue`):
  batch vs. external sbatch generation; the `$SLURM_JOB_ID` fallback; the `_kill`
  batch/external split; the cached-job liveness path; and the runner (client run
  as argv with `MONARCH_BATCH_JOB`, worker teardown, and the `kill` fallback when
  `terminate()` times out).
- **End to end** on a live SLURM cluster (25.05) driving a real multi-actor
  workload:
  - single node (controller colocated): COMPLETED, clean teardown;
  - two nodes (workers split across nodes): COMPLETED -- correct
    hostname-to-mesh distribution, cross-node state transfer, and no dangling
    allocation.

Reviewed By: zdevito

Differential Revision: D109482857

Pulled By: JenniferWang


zdevito (approved):
This looks good! Can you ghimport it to fbcode and then Ill stamp it there and we can land. I am not sure if we have automated tests of SlurmJob, can you also quickly check that interactive mode is still running as well?

JenniferWang:
> This looks good! Can you ghimport it to fbcode and then Ill stamp it there and we can land. I am not sure if we have automated tests of SlurmJob, can you also quickly check that interactive mode is still running as well?

YEs. I've verified the interactive mode is still running well.

PR #4292, opened 2026-06-24
Fix incorrect CUdeviceptr init

alugorey:
`CUdeviceptr` is an integer on CUDA but a raw pointer (`hipDeviceptr_t`) on ROCm, so initialize portably with `zeroed` (0 / null)


alugorey:
This has been imported here; https://github.com/meta-pytorch/monarch/pull/4341

will close this once that one gets merged

PR #4304, opened 2026-06-26
Add auto-detect for RoCE v2 GID for mlx5 NICs in MlxDevice

mkuchnik:
rdma-core's default gid_index (3) is not the routable GID on RoCE v2 fabrics, and mlx5 does no auto-selection, so QPs come up on an unroutable GID. This diff adds autoselection logic for MlxDevice:
  1. MONARCH_IB_GID_INDEX, if set (as an override);
  2. else auto-detect the first routable RoCE v2 GID from sysfs (gid_attrs/types/<i> == "RoCE v2" and gids/<i> not link-local / all-zero);
  3. else fall back to gid_index 3 (prior InfiniBand-mode NICs behavior).

The result is cached process-wide.
Tested on a GB300 node with ConnectX-8 SuperNIC, where the routable GID is index 7.


monarch_rdma/src/backend/ibverbs/mlx_device.rs
```diff
@@ -42,7 +43,9 @@ impl IbvDeviceImpl for MlxDevice {
         queried && attr.vendor_id == MELLANOX_VENDOR_ID
     }
 
-    fn apply_config_defaults(_config: &mut IbvConfig) {}
+    fn apply_config_defaults(config: &mut IbvConfig) {
+        config.gid_index = roce_gid::select_roce_v2_gid_index();
```
samlurye: I would prefer the selection function to be optional, and if it returns `None`, just don't do any override.

monarch_rdma/src/backend/ibverbs/roce_gid.rs
```diff
@@ -0,0 +1,237 @@
...
+//! keeps rdma-core's default, leaving InfiniBand-mode NICs on their prior
+//! behavior.
+
+use std::sync::OnceLock;
+
+/// rdma-core's default GID index, and the value this backend previously used for
+/// InfiniBand-mode mlx5 NICs. Kept as the backward-compatible fallback when
+/// neither an explicit override nor sysfs auto-detection yields a RoCE v2 GID
+/// (e.g. an InfiniBand-mode mlx5 NIC, which has no `RoCE v2` GID entry).
+const DEFAULT_GID_INDEX: u8 = 3;
```
samlurye: Let's avoid having the hardcoded default here, and just return `None` as mentioned above.

monarch_rdma/src/backend/ibverbs/roce_gid.rs
```diff
@@ -0,0 +1,237 @@
...
+                );
+                DEFAULT_GID_INDEX
+            }
+        }
+    })
+}
+
+/// `MONARCH_IB_GID_INDEX` parsed as a `u8`, if set and valid.
+fn env_gid_index() -> Option<u8> {
+    std::env::var("MONARCH_IB_GID_INDEX")
```
samlurye: The canonical way to define an env var option in monarch is using a hyperactor config attribute; the options for RDMA are defined [here](https://github.com/meta-pytorch/monarch/blob/main/monarch_rdma/src/config.rs#L17). Please call it `RDMA_IBV_DEFAULT_GID_INDEX`, and incorporate it into the code base by setting the value in [`IbvConfig`'s default implementation](https://github.com/meta-pytorch/monarch/blob/main/monarch_rdma/src/backend/ibverbs/primitives.rs#L194).

monarch_rdma/src/backend/ibverbs/roce_gid.rs
```diff
@@ -0,0 +1,237 @@
+/*
```
samlurye: Instead of having this logic in its own module, let's add it on the [`primitives::Gid` type](https://github.com/meta-pytorch/monarch/blob/main/monarch_rdma/src/backend/ibverbs/primitives.rs#L53). We can add a new type:
```rust
enum GidType {
  Loopback,
  LinkLocal,
  SiteLocal,
  Global  
}

impl GidType {
  fn type_of(maybe_gid: &str) -> Option<Self> { ... }
}
```
Then on `Gid`, we can make the logic a lot simpler:
```rust
impl Gid {
  fn index_of<I: IbvDeviceImpl>(gid_type: GidType) -> Option<u8> {
    // List every device of the specified type
    for device_info in I::list() {
      // `device_info: IbvDeviceInfo` already contains the list of ports
      // for the device of interest, and each port struct already contains
      // the gid as a string
    }
  }
}
``` 
So for `MlxDevice::apply_config_defaults`, you would just call `Gid::index_of::<MlxDevice>(GidType::Global)`.

samlurye (changes requested):

monarch_rdma/src/backend/ibverbs/roce_gid.rs
mkuchnik: Take a look. The downside of spreading autoconfig between 1) env 2) autoconfig 3) default is there is an unclear ordering of priority. For example, autoconfig is applied after env and default, so the priority order seems 2, 1, 3, which may be unexpected.

monarch_rdma/src/config.rs
```diff
@@ -79,4 +79,14 @@ declare_attrs! {
...
+
+    /// Default GID index for mlx5 NICs (else 3). `MlxDevice::apply_config_defaults`
+    /// overrides it with the auto-detected routable RoCE v2 GID when one is found,
+    /// so this value applies when auto-detection finds nothing (e.g. an
+    /// InfiniBand-mode mlx5 NIC).
+    @meta(CONFIG = ConfigAttr::new(
+        Some("MONARCH_RDMA_IBV_DEFAULT_GID_INDEX".to_string()),
+        Some("rdma_ibv_default_gid_index".to_string()),
+    ))
+    pub attr RDMA_IBV_DEFAULT_GID_INDEX: Option<u8> = None;
```
samlurye, line 91: Let's make this non-optional with a default of 0. The comment shouldn't mention mlx5 or any specific hardware.

monarch_rdma/src/backend/ibverbs/primitives.rs
```diff
@@ -64,7 +65,160 @@ impl Gid {
     fn interface_id(&self) -> u64 {
         u64::from_be_bytes(self.raw[8..].try_into().unwrap())
     }
+
+    /// The GID-table index of the first RoCE v2 GID of `gid_type` on any device
+    /// of backend `I`, or `None` if none is present.
+    ///
+    /// Reads the kernel's GID table from sysfs
+    /// (`/sys/class/infiniband/<dev>/ports/<p>/gid_attrs/types/<i>` and
+    /// `gids/<i>`) rather than `ibv_query_gid*`: on some mlx5 / RoCE stacks the
```
samlurye, line 74: A few things:

1. Comments in non-hardware-specific files shouldn't be mentioning specific hardware that uses it. 
2. Are you actually in practice running into cases where `ibv_query_gid` returns  a bad value?
3. If `device_info` doesn't contain the right thing, then we should update `IbvDeviceInfo` so that it *does* contain the right thing. Instead of having each `IbvPort` contain only a single gid, let's give it the full list of gids, read from sysfs (filtering out non-roce-v2 gids), and fill it in inside `query_device_info`. Then `Gid::index_of` can just read from `IbvDeviceInfo` without having to go through sysfs or any additional `ibv_query_gid` calls.
4. Please tell the LLM to be more concise in its comments.

samlurye (changes requested):

PR #4307, opened 2026-06-26
Make actor runtime device-agnostic (support XPU alongside CUDA)

songhappy:
Summary
 Remove three CUDA-only assumptions from the actor runtime (Layer 1) so Monarch works on any
 accelerator supported by torch.accelerator without adding torch as a dependency.

  - `device_utils.py`: `_local_device_count()` also honors `ZE_AFFINITY_MASK`.
  - `proc_mesh.py`: the accelerator env-var snapshot/warning helpers cover
    `ZE_AFFINITY_MASK`, `ZE_ENABLE_PCI_ID_DEVICE_ORDER` and
    `PYTORCH_XPU_ALLOC_CONF`; the initialized-check dispatches to `torch.xpu` or
    `torch.cuda`.

 No Layer 2 changes. CUDA remains the default. 5 files, +600/-31.

  CUDA remains the default. Torch stays optional: no unconditional imports, and
  the env-var / `/dev/nvidia*` probes run before torch is consulted, so torch-free
  behavior is unchanged.

  ## Test plan

  `pytest python/tests/test_xpu.py -v` — 16 tests. 5 need no hardware and cover
  the XPU, CUDA and no-torch branches; the rest need an XPU.
  - Intel Data Center GPU Max 1550, 12 tiles, torch 2.12.0+xpu: 16 passed.

### DIFF
--- a/python/monarch/_src/actor/device_utils.py
+++ b/python/monarch/_src/actor/device_utils.py
@@ -6,15 +6,8 @@
 
 # pyre-strict
 
-import os
-import re
-from pathlib import Path
+import torch
 
 
 def _local_device_count() -> int:
-    if "CUDA_VISIBLE_DEVICES" in os.environ:
-        return len(os.environ["CUDA_VISIBLE_DEVICES"].split(","))
-    dev_path = Path("/dev")
-    pattern = re.compile(r"nvidia\d+$")
-    nvidia_devices = [dev for dev in dev_path.iterdir() if pattern.match(dev.name)]
-    return len(nvidia_devices)
+    return torch.accelerator.device_count()
--- a/python/monarch/_src/actor/proc_mesh.py
+++ b/python/monarch/_src/actor/proc_mesh.py
@@ -98,55 +98,84 @@ _COMMON_CUDA_ENV_VARS: Tuple[str, ...] = (
     "PYTORCH_CUDA_ALLOC_CONF",
 )
 
+_COMMON_XPU_ENV_VARS: Tuple[str, ...] = (
+    "ZE_AFFINITY_MASK",
+    "ZE_ENABLE_PCI_ID_DEVICE_ORDER",
+    "PYTORCH_XPU_ALLOC_CONF",
+)
+
+
+def _get_accelerator_env_vars() -> Tuple[str, ...]:
+    """Return the relevant env vars based on the active accelerator."""
+    import torch
+
+    if torch.accelerator.current_accelerator() == "xpu":
+        return _COMMON_XPU_ENV_VARS + _COMMON_CUDA_ENV_VARS
+    return _COMMON_CUDA_ENV_VARS
+
 
 T = TypeVar("T")
 TActor = TypeVar("TActor", bound=Actor)
 
 
-def _cuda_env_snapshot() -> Dict[str, Optional[str]]:
-    return {key: os.environ.get(key) for key in _COMMON_CUDA_ENV_VARS}
+def _accel_env_snapshot() -> Dict[str, Optional[str]]:
+    env_vars = _get_accelerator_env_vars()
+    return {key: os.environ.get(key) for key in env_vars}
 
 
-def _torch_cuda_already_initialized() -> bool:
+def _torch_accelerator_already_initialized() -> bool:
+    import torch
+
+    accel = torch.accelerator.current_accelerator()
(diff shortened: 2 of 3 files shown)


songhappy:
@shayne-fletcher would you please kindly have somebody review this PR?

dulinriley:
I'm sorry I didn't see this earlier. I am not a good reviewer for GPU stuff. @samlurye can help here

songhappy:
@samlurye Please kindly review it

python/monarch/_src/job/job.py
```diff
@@ -29,7 +29,30 @@
 from monarch._src.job._batch_env import in_batch_job, MONARCH_BATCH_JOB_ENV
 from monarch._src.job.job_sidecar import stop_job_sidecar
 from monarch._src.job.mount_config import Mounts
-from monarch._src.job.telemetry_config import TelemetryConfig
+
+# `telemetry_config` transitively imports the SQL-backed
+# `monarch_distributed_telemetry` Rust submodule, which is only built when
+# MONARCH_ENABLE_SQL_TELEMETRY=1. Skip-and-stub `TelemetryConfig` on
+# actors-only builds (e.g. XPU) where the submodule is absent.
```
dulinriley: This seems weird, we should be including telemetry even in a non-tensor engine build.
cc @thedavekwon is this a build error or just a misunderstanding?

setup.py
```diff
@@ -381,7 +381,9 @@ def create_cpp_extension(
 rust_extensions = []
 
 # Main Python extension
-rust_features = ["extension-module", "distributed_sql_telemetry"]
+rust_features = ["extension-module"]
+if build_tensor_engine:
+    rust_features.append("distributed_sql_telemetry")
```
dulinriley: This is wrong, we should always build the distributed_sql_telemetry

python/monarch/_src/job/job.py
songhappy: You're right -- reverted. `distributed_sql_telemetry` is always built now, so the try/except and stub are removed. Direct import restored.

setup.py
songhappy: Fixed -- `distributed_sql_telemetry` is back to always-on, not gated behind `build_tensor_engine`.

songhappy:
@dulinriley could you please enable the workflows?

songhappy:
Please approve and merge if it's all OK, I will be on travel for 4 weeks from 8.10. Please do not make it stale.

samlurye:
> Please approve and merge if it's all OK, I will be on travel for 4 weeks from 8.10. Please do not make it stale.

@songhappy Monarch intentionally does not take `torch` as a dependency, and so we can't have any module-level or unconditional imports of `torch`. So there are a few things this needs in order to be accepted:
- Do not overwrite any existing logic that did not previously depend on torch in order to now depend on torch
- To the extent possible, any new logic you add into monarch core should not require torch
- If you absolutely must rely on torch, use the existing fallible import pattern (e.g., `torch_mod = sys.modules.get("torch")`, then check if `torch_mod` is not `None`)

songhappy:
@samlurye Reworked — torch is no longer required anywhere in this change.

  Both new helpers reach torch via `sys.modules` instead of importing it:

  ```python
  torch_mod = sys.modules.get("torch")
  accelerator = getattr(torch_mod, "accelerator", None)
  if accelerator is None:
      return None
 ```
  I also reverted the part where _local_device_count() replaced the /dev/nvidia* scan with
  torch. The env-var and /dev/nvidia* probes run first, so behavior without torch is unchanged.

  setup.py and job.py are dropped per @dulinriley. Now 3 files: device_utils.py, proc_mesh.py,
  test_xpu.py.

  Tested on an Intel Data Center GPU Max node (12 tiles, torch 2.12.0+xpu): 16 passed.

> @songhappy Monarch intentionally does not take `torch` as a dependency, and so we can't have any module-level or unconditional imports of `torch`. So there are a few things this needs in order to be accepted:
> 
> * Do not overwrite any existing logic that did not previously depend on torch in order to now depend on torch
> * To the extent possible, any new logic you add into monarch core should not require torch
> * If you absolutely must rely on torch, use the existing fallible import pattern (e.g., `torch_mod = sys.modules.get("torch")`, then check if `torch_mod` is not `None`)

PR #4341, opened 2026-07-01
[ROCm] Re-order torch import to avoid race with rocprofiler

alugorey:
monarch's native extension links the system libamdhip64 while torch bundles its own; when monarch._rust_bindings was imported before torch, the system HIP runtime won rocprofiler-register and   caused `hipApiName has non-null function pointer 1 despite this being the first instance of the library being copies`. In this PR we ensure torch is imported first. The comments above the original lines insinuated this was supposed to already happen, but the code was written in the reverse.

Additionally, we fix a CUdeviceptr bug:
CUdeviceptr is an integer on CUDA but a raw pointer (hipDeviceptr_t) on ROCm, so initialize portably with zeroed (0 / null)

### DIFF
--- a/python/monarch/__init__.py
+++ b/python/monarch/__init__.py
@@ -9,47 +9,67 @@
 from importlib import import_module as _import_module
 from typing import TYPE_CHECKING
 
-# Opt-in torch preload, BEFORE monarch._rust_bindings.
-#
-# On ROCm, monarch's native extension links the system libamdhip64 while torch
-# bundles its own copy. If monarch._rust_bindings loads first, the system HIP
-# runtime wins rocprofiler-register and the first HIP runtime call afterward
-# aborts (hip.cpp:512 "hipApiName has non-null function pointer ... first
-# instance of the library being copied", SIGABRT). Importing torch first makes
-# torch's runtime win.
-#
-# This is opt-in via MONARCH_PRELOAD_TORCH so monarch stays importable -- and
-# torch-free -- for non-torch frameworks. _get_bootstrap_args() sets it on
-# spawned procs only when the launching process itself uses torch, so CPU /
-# no-torch workloads never pull torch in.
 import os as _os
-
-if _os.environ.get("MONARCH_PRELOAD_TORCH") == "1":
-    try:
-        import torch  # @manual  # noqa: F401
-    except ImportError:
-        pass
-
-# Import before monarch to pre-load torch DSOs as, in exploded wheel flows,
-# our RPATHs won't correctly find them.
 import sys as _sys
 
-# Record whether torch's HIP runtime was already loaded when we load our native
-# extension. On ROCm, if our extension (which links the system libamdhip64)
-# loads before torch's bundled copy, the first HIP call aborts
-# (hip.cpp:512 "hipApiName ..."). The RDMA path checks this flag to raise a
-# clear, actionable error instead of a hard SIGABRT. See
+
+def _preload_torch_hip_runtime() -> bool:
+    """Preload torch's bundled ROCm HIP shared library before our native
+    extension loads.
+
+    On ROCm, ``monarch._rust_bindings`` links the *system* ``libamdhip64`` while
+    torch ships its *own* copy. Whichever HIP runtime initializes first wins
+    ``rocprofiler-register``; if the system copy wins, the first HIP call made
+    afterward aborts the process
+    (``hip.cpp:512 "hipApiName has non-null function pointer ..."``). Loading
+    torch's ``libamdhip64`` here makes torch's runtime win regardless of the
+    order in which the user imports torch and monarch.
+
+    This is cheap (~tens of ms, just a dlopen) and does NOT import the torch
+    Python module, so it never pulls torch into non-torch workloads. It is a
+    no-op when torch is not installed or is a CUDA/CPU build (the HIP DSO is
+    absent). Set ``MONARCH_PRELOAD_TORCH=0`` to skip it.
+
+    Returns True when torch's HIP runtime is present (already imported, or loaded
+    here) so callers can record that the load ordering is safe.
+    """
+    if _os.environ.get("MONARCH_PRELOAD_TORCH") == "0":
+        return "torch" in _sys.modules
+    if "torch" in _sys.modules:
+        return True  # torch already imported: its HIP runtime is already loaded
+    try:
+        import ctypes
+        import importlib.util
+
+        spec = importlib.util.find_spec("torch")
+        if spec is None or spec.origin is None:
+            return False
+        lib = _os.path.join(_os.path.dirname(spec.origin), "lib", "libamdhip64.so")
+        if not _os.path.exists(lib):
+            return False  # CUDA/CPU torch build: no HIP runtime to preload
+        ctypes.CDLL(lib, mode=ctypes.RTLD_GLOBAL)
+        return True
+    except Exception:
+        return False
+
+
+# Preload torch's HIP runtime (ROCm) BEFORE importing our native extension so
+# torch's copy wins rocprofiler-register; otherwise the first HIP call aborts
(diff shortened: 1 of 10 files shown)


alugorey:
FYI @samlurye @dulinriley @jithunnair-amd

alugorey:
Hi @samlurye , Could i get you to look at the most recent commit in this PR? the first one fixed the RDMA test failure but it spawned a different failure in test_rdma_buffer_cpu_memoryview which complained about torch being present in sys.modules which broke the contract in the test. the second commit fixed the issue but i want to make sure I don't go down the wrong path here.

alugorey:
@samlurye @dulinriley @mreso Finally got ROCm CI green with this one. Could you please take a look? I did what i had to do to get CI green but i'm sure there are things i didn't take into account or violated some contracts that we'd like to preserve so please look over. We can use this as the jumping off point at least.

python/monarch/_src/rdma/rdma.py
```diff
@@ -221,9 +221,49 @@ def _evict_local_memory(key: Tuple[int, int, int], ref: "weakref.ref") -> None:
...
+        with open("/proc/self/maps") as f:
+            maps = f.read()
+    except OSError:
+        return
+    import re
+
+    libs = set(re.findall(r"/\S*libamdhip64\.so[.0-9]*", maps))
+    if len(libs) < 2:
+        return  # single HIP runtime: safe
+    raise RuntimeError(
```
dulinriley: Maybe print all the library paths found in this case?

dulinriley (approved):

dulinriley:
@alugorey I had to make a slight patch to fix formatting and a real bug related to device masking. Please update this PR with this patch file:
(use `git am` to apply because this was created with `format-patch`)
[fix-rocm.patch](https://github.com/user-attachments/files/30363198/fix-rocm.patch)

alugorey:
> @alugorey I had to make a slight patch to fix formatting and a real bug related to device masking. Please update this PR with this patch file: (use `git am` to apply because this was created with `format-patch`) [fix-rocm.patch](https://github.com/user-attachments/files/30363198/fix-rocm.patch)

@dulinriley Patch applied

samlurye:
@alugorey Without this fix, does the `hipApiName has non-null function pointer` error happen always on a call from monarch into hip, or can it also happen on a torch api call? In other words, if you first import `monarch._rust_bindings`, then import `torch` and do something with a hip tensor, what happens?

dulinriley (approved):

python/monarch/__init__.py
```diff
@@ -9,15 +9,67 @@
 from importlib import import_module as _import_module
 from typing import TYPE_CHECKING
 
-# Import before monarch to pre-load torch DSOs as, in exploded wheel flows,
-# our RPATHs won't correctly find them.
+import os as _os
```
dulinriley: Are you sure you applied the patch? The internal formatter is complaining about these import positions

dulinriley (changes requested):

python/monarch/__init__.py
dulinriley: Sorry about this, I wish I could just apply the formatting changes directly, but the internal code and the PR need to be in sync and I'm not allowed to update your branch on your fork. (if you make a branch directly on meta-pytorch/monarch I could edit that directly).

alugorey: ah right. so the patch didn't actually seem to apply cleanly. much of the patch had changes that were already in this PR and caused collisions. Claude also said that the patch was a "Mercurial" export which didn't make sense to me. I went ahead and manually added the changes so i may have missed something. Can you re-inspect the patch and make sure it was built right on top of this PR? I don't believe it should have any changes in the diff that are already present in this PR. Also not sure why claude said it was a "mercurial" export or what that even means. in any case, the patch did not apply cleanly using `git am` which is why i pivoted the approach. I also don't think i have the permissions to make a branch directly on the meta-pytorch/monarch repo.

alugorey:
> @alugorey Without this fix, does the `hipApiName has non-null function pointer` error happen always on a call from monarch into hip, or can it also happen on a torch api call? In other words, if you first import `monarch._rust_bindings`, then import `torch` and do something with a hip tensor, what happens?

@samlurye it's less about monarch to hip and torch to hip. the issue happens because there are two copies of libamdhip64 in our case. one on the system (usually in /opt/rocm) and then one bundled with torch. So before this change it became a race to register the hip runtimes with rocprofiler. Whichever runtime registered with rocprofiler **second** would abort. the problem requires both torch and monarch to use hip in the same process (like in the unit tests).

import `monarch._rust_bindings`, then import `torch` and do something with a hip tensor, what happens
this would **not** crash as the system copy is loaded but never initialized. the order dependent symptom would look as follows: 
- monarch-second -> hipApiName SIGABRT (rdmaxcel path)
- torch-second -> torch "No CUDA GPUs are available"

alugorey:
@dulinriley i added you as a collaborator to my fork. perhaps you can re-try adding your changes, if that makes it easier?

samlurye:
> import monarch._rust_bindings, then import torch and do something with a hip tensor, what happens
this would not crash as the system copy is loaded but never initialized. the order dependent symptom would look as follows

@alugorey Instead of relying on import order for correctness (like in this PR), which won't work if someone pulls in a python module that loads hip but that isn't torch, I'm wondering if a more sustainable solution would be to just ensure that monarch never loads the system hip.

Currently, there are two places in monarch's build that link against libamdhip64:
- The `nccl-sys` crate, which belongs to the tensor engine feature and which no one is using.
- The `rdmaxcel-sys` crate, but the symbols requiring libamdhip64 linkage are used only in kernel-launched RDMA functionality, which also currently isn't being used (and may be broken anyway).

So my proposal would be to do the following:
- Decouple the RDMA build from the `USE_TENSOR_ENGINE` flag. Making RDMA usable without the tensor engine will mean that `nccl-sys` won't be built, so that module will never load hip.
- Gate kernel-launched RDMA behind a new flag, default off. Then for common usage, the RDMA build won't need to link against hip and so monarch will never load hip.

What do you think?

python/monarch/__init__.py
dulinriley: Oh haha mercurial is the git alternative we use inside Meta: https://www.mercurial-scm.org/

That's interesting it rejected the patch because I used git to apply it as a test and it seemed to apply cleanly. But no matter, I can just checkout your branch now and directly apply the formatter.

dulinriley:
You should work with @samlurye to address the comments. Once that's done either he or I can apply any formatting changes needed and re-import

alugorey:
> > import monarch._rust_bindings, then import torch and do something with a hip tensor, what happens
> > this would not crash as the system copy is loaded but never initialized. the order dependent symptom would look as follows
> 
> @alugorey Instead of relying on import order for correctness (like in this PR), which won't work if someone pulls in a python module that loads hip but that isn't torch, I'm wondering if a more sustainable solution would be to just ensure that monarch never loads the system hip.
> 
> Currently, there are two places in monarch's build that link against libamdhip64:
> 
> * The `nccl-sys` crate, which belongs to the tensor engine feature and which no one is using.
> * The `rdmaxcel-sys` crate, but the symbols requiring libamdhip64 linkage are used only in kernel-launched RDMA functionality, which also currently isn't being used (and may be broken anyway).
> 
> So my proposal would be to do the following:
> 
> * Decouple the RDMA build from the `USE_TENSOR_ENGINE` flag. Making RDMA usable without the tensor engine will mean that `nccl-sys` won't be built, so that module will never load hip.
> * Gate kernel-launched RDMA behind a new flag, default off. Then for common usage, the RDMA build won't need to link against hip and so monarch will never load hip.
> 
> What do you think?

I'm totally okay with that! honestly, it was kind of a pain in the butt trying to handle all the ordering edge cases, so if we can sever the need to load libamdhip64 altogether and allow pytorch to just manage that however it wants then that's a much cleaner solution. Give me a day or so to work on that and i'll circle back.

PR #4415, opened 2026-07-15
Fix macOS telemetry thread ID build

dulinriley:
Summary:
There was a macOS build failure in `hyperactor_telemetry` because the non-Linux `get_thread_info()` path calls `ThreadId::as_u64()`, which is gated behind `thread_id_value` on the pinned nightly.

This change enables `thread_id_value` only on macOS with `cfg_attr(target_os = "macos", ...)`, so Linux does not enable an unused unstable feature.

Test Plan:
- `RUSTFLAGS="--cfg tracing_unstable" cargo check -p hyperactor_telemetry -j2`
- `RUSTFLAGS="--cfg tracing_unstable" cargo build -p hyperactor_telemetry -j2`

### DIFF
--- a/hyperactor_telemetry/src/lib.rs
+++ b/hyperactor_telemetry/src/lib.rs
@@ -7,6 +7,7 @@
  */
 
 #![allow(internal_features)]
+#![cfg_attr(target_os = "macos", feature(thread_id_value))]
 #![feature(sync_unsafe_cell)]
 #![feature(mpmc_channel)]
 #![feature(formatting_options)]


dulinriley:
Additional macOS wheel-build evidence:

- `USE_TENSOR_ENGINE=1 UV_PYTHON=.venv/bin/python CARGO_BUILD_JOBS=2 uv build --no-build-isolation --wheel`
- Produced `dist/torchmonarch-0.7.0.dev0-cp312-cp312-macosx_26_0_arm64.whl`

This built the macOS arm64 wheel locally with tensor_engine CPU-only. The wheel path compiled the main `_rust_bindings` extension and the `hyperactor_mesh_admin_tui` binary; `hyperactor_telemetry` compiled successfully in both graphs, with no `thread_id_value` error.

Local caveats: this machine does not have `npm`, so setup.py used its existing warning path and skipped the dashboard frontend build; the source tree already had an up-to-date `monarch.common._C` extension, so setup.py copied that cached `.so` instead of rebuilding it.

PR #4619, opened 2026-08-05
Make a worker's advertised address and its peers' dial address the same string

JenniferWang:
## Problem

On CoreWeave GB300 each node's `/etc/hosts` adds an IPv6 GUA for its own name while cluster DNS publishes only an A record, so a worker serves on `tcp://[GUA]:PORT` while the controller dials `tcp://IPv4:PORT`. As a result, multi-node attach times out with `MESH_ATTACH_CONFIG_TIMEOUT` on clusters where a node's own name resolves differently on the node than it does from its peers. Single node works by accident: the controller resolves that node through the same `/etc/hosts` the worker did.

The two addresses must agree **byte for byte**, not merely be mutually reachable. A worker's host identity is the `service` proc derived from the address it *advertises*, and `Proc::is_local_delivery_target_at` compares the terminal channel address for that proc — so a mismatch fails delivery with `ttl expired` after the TCP connection has already succeeded. That rules out the fix this looks like it wants: widening the bind (the `advertise@bind` alias form with a wildcard bind, as `KubernetesJob` uses) fixes reachability, which was never the problem.

## Fix 
Both sides now apply one deterministic rule to whatever their resolver returned — drop loopback and link-local, prefer IPv4, break ties by address — landing in `monarch/tools/network.py` as `get_consistent_sockaddr`. Full rationale, including why this is a new function rather than an `AddrType` variant and why the metatls branches keep `getfqdn()`, is in the commit message.

## Call sites

- `_src/job/_slurm_batch.py` + `_src/job/slurm.py` — `_WORKER_BOOTSTRAP` and `SlurmJob._state` derived the address independently and hoped they matched; they now share `worker_address`, so they agree by construction rather than by convention across two files.
- `_src/spmd/host_mesh.py` — `_worker_address` published `tcp://{getfqdn()}:{port}` into the store. Publishing a hostname doesn't help, because `ChannelAddr::Tcp` holds a *resolved* `SocketAddr`; publishing a literal does, because it resolves to itself.

## Verification

Resolving one node name from both sides on the CoreWeave GB300 cluster (fair-cw-use2-3):

```
node itself   raw: AF_INET6 2a03:83e4:4027:8adc::be0d, AF_INET 10.133.110.19
login side    raw: AF_INET 10.133.110.19

get_consistent_sockaddr   both sides -> 10.133.110.19:22222        (agree)
get_sockaddr (IPv6-first) node       -> [2a03:...::be0d]:22222
                          login      -> 10.133.110.19:22222        (disagree)
```

End to end — two-node `SlurmJob` in batch mode, same client and allocation shape, swapping only the monarch Python:

| | result |
|---|---|
| before | **FAIL** — `config push failed during attach on 1 host(s): tcp:10.133.173.11:22222 reply timed out after MESH_ATTACH_CONFIG_TIMEOUT` |
| after | **PASS** — actors ran on both `g3-154-001` and `g3-154-005` |

One of two hosts failing is the signature this predicts: the head node resolves its own name the same way its worker did, so that half agrees by accident, while the remote node's name comes from DNS as IPv4 and its worker had bound the GUA.

Caveat on scope: the end-to-end runs overlaid this branch's Python onto a prebuilt `_rust_bindings` (2026-07-29) rather than a from-source build of current `main`. All four changed files are pure Python, so this exercises the actual changed code, but not a full rebuild.

## Notes for reviewers

- **Tie-break by address, not resolver order.** `getaddrinfo` ordering is host-configurable (RFC 6724, `gai.conf`), so a name with several A records would otherwise still resolve differently per host — the exact failure this exists to prevent. No-op on this cluster (one A + one AAAA per node).
- **The rule absorbs one asymmetry, not every one.** An extra IPv6 visible only to the node can't change the choice; an extra IPv4 would. That's the case in the field, and the docstring says so rather than overclaiming.
- **`test_external_controller_mode_has_no_client`** now matches `-m monarch._src.job._slurm_batch` instead of the bare module name, because the worker bootstrap legitimately imports `worker_address` from it.

## Out of scope

The worker could publish the address it actually bound and the controller consume it through the `Future[str]` support `attach_to_workers` already documents, removing independent resolution entirely rather than making two resolutions agree. The SLURM worker could also name itself with `$SLURMD_NODENAME`, which matches squeue's node name by definition even when a site sets `NodeName != NodeHostname`. Separately, the `tcp://*` guard in `run_worker_loop_forever` is a substring test over the whole address, so it also rejects a wildcard in the `bind_to` half of an alias, where one is legitimate.

### DIFF
--- a/python/monarch/tools/network.py
+++ b/python/monarch/tools/network.py
@@ -54,6 +54,75 @@ def get_sockaddr(hostname: str, port: int) -> str:
     )
 
 
+def get_consistent_sockaddr(hostname: str, port: int) -> str:
+    """Returns the socket address of `hostname:port` that *every* host resolves alike.
+
+    Same return shape as `get_sockaddr`: `{ipv4}:{port}` or `[{ipv6}]:{port}`.
+
+    Use this rather than `get_sockaddr` when the address is a shared *identity*:
+    one host advertises where it can be reached and another dials it, and the
+    two strings must be equal, not merely both reachable. `ChannelAddr::Tcp`
+    holds a resolved socket address and a worker's host identity is derived from
+    the address it advertises, so a peer that resolves the same name differently
+    connects and then fails delivery with "ttl expired".
+
+    Resolvers do differ: a node's own name goes through its `/etc/hosts`, which
+    can publish families that DNS does not -- on CoreWeave GB300 each node's
+    `/etc/hosts` adds an IPv6 GUA for its own name while cluster DNS serves only
+    an A record. So drop loopback and link-local (the routability rule
+    `ChannelAddr::any` uses), prefer IPv4, and break ties by address. IPv4-first
+    is what makes the side that sees the extra IPv6 agree with the side that
+    does not; `AddrType.Default` is IPv6-first and would not, and pinning
+    `AddrType.IPv4` would agree but break IPv6-only clusters.
+
+    Raises a `RuntimeError` if `hostname` resolves to no routable address.
+    """
+    candidates: list[tuple[int, ipaddress.IPv4Address | ipaddress.IPv6Address]] = []
+    # patternlint-disable-next-line python-dns-deps (only used for oss)
+    for family, _, _, _, sockaddr in socket.getaddrinfo(
+        hostname, port, type=socket.SOCK_STREAM
+    ):
+        # Strip any scope id ("fe80::1%eth0"), which ip_address rejects.
+        addr = str(sockaddr[0]).partition("%")[0]
+        ipaddr = ipaddress.ip_address(addr)
+        # Loopback is dropped as well as link-local (unlike `_resolve_ipaddr`):
+        # every host resolves it alike, but no peer can reach it.
+        if ipaddr.is_loopback or ipaddr.is_link_local:
+            logger.info(
+                "skipping unroutable address `%s` for `%s:%d`"
+                " (loopback and link-local addresses cannot be dialed by a peer)",
+                addr,
+                hostname,
+                port,
+            )
+            continue
+        candidates.append((family, ipaddr))
+
+    if not candidates:
+        raise RuntimeError(
+            f"Unable to resolve `{hostname}` to a routable (non-loopback,"
+            " non-link-local) address that can bind TCP socket."
+            " Check the network configuration on the host."
+        )
+
+    # Least address within the preferred family, not the first one the resolver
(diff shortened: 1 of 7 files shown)


python/monarch/tools/network.py
```diff
@@ -54,6 +54,75 @@ def resolve_sockaddr(family: socket.AddressFamily) -> Optional[str]:
     )
 
 
+def get_consistent_sockaddr(hostname: str, port: int) -> str:
```
dulinriley, line 57: nit: let's use the `ipaddress` object as the return type instead of the string.

python/monarch/tools/network.py
```diff
@@ -54,6 +54,75 @@ def resolve_sockaddr(family: socket.AddressFamily) -> Optional[str]:
...
+    `AddrType.IPv4` would agree but break IPv6-only clusters.
+
+    Raises a `RuntimeError` if `hostname` resolves to no routable address.
+    """
+    candidates: list[tuple[int, ipaddress.IPv4Address | ipaddress.IPv6Address]] = []
+    # patternlint-disable-next-line python-dns-deps (only used for oss)
+    for family, _, _, _, sockaddr in socket.getaddrinfo(
+        hostname, port, type=socket.SOCK_STREAM
+    ):
+        # Strip any scope id ("fe80::1%eth0"), which ip_address rejects.
```
dulinriley, line 85: If you are on python 3.9+ ip_address already removes the scope part.

python/monarch/tools/network.py
```diff
@@ -54,6 +54,75 @@ def resolve_sockaddr(family: socket.AddressFamily) -> Optional[str]:
...
+            f"Unable to resolve `{hostname}` to a routable (non-loopback,"
+            " non-link-local) address that can bind TCP socket."
+            " Check the network configuration on the host."
+        )
+
+    # Least address within the preferred family, not the first one the resolver
+    # happened to list: that ordering is host-specific (RFC 6724, gai.conf) and
+    # a name with several A records would otherwise resolve differently per host.
+    family, ipaddr = min(
+        candidates, key=lambda c: (c[0] != socket.AF_INET, c[1].packed)
```
dulinriley, line 112: Why do we prefer IPv4 here? And why do we sort by the IP address? This seems like it would prefer an address starting with 10.* over an address with 12.*, when both might be equally valid?

dulinriley (changes requested):

PR #4653, opened 2026-08-11
Add out-of-cluster mode to `SlurmJob` (#4653)

dulinriley:
Summary:

Add `out_of_cluster` and `attach_to` configuration to `SlurmJob`. In
out-of-cluster mode, attach the client gateway through the first allocated
worker before materializing any host mesh. An explicit `attach_to` address can
override that worker address, including when a caller provides an SSH-forwarded
local endpoint.

Include both options in cached-job compatibility checks. Add regression tests
that exercise the public `state()` lifecycle and verify gateway selection,
attachment ordering, explicit overrides, and unchanged in-cluster behavior.

Differential Revision: D115629683


JenniferWang (approved):

Issue #4749, opened 2026-08-25
`shutdown_context()` can return while Python actor event-loop threads are still running


shayne-fletcher:
## Problem

each `PythonActor` owns a dedicated daemon thread running an asyncio event loop. normal actor cleanup schedules that loop to stop, but it does not wait for the thread to exit. the root client is also a `PythonActor`, but its custom message loop bypasses the normal actor teardown path and can report `Stopped` without running actor cleanup.
as a result, `shutdown_context().get()` can complete while actor event-loop threads are still running. when CPython begins interpreter finalization, one of those daemon threads may attempt to acquire the GIL. on CPython 3.12 and earlier, the thread is force-exited; if that unwind crosses a PyO3 callback boundary, the process aborts with:
```text
FATAL: exception not rethrown
```
the failure is attributed to whichever test subprocess happens to be exiting at the time, so the apparent failing test varies between runs.
## History
- PR #1985 made the root client a real `PythonActor`, giving it an actor-local event loop, and made actor-loop threads daemon threads so an unstopped loop would not hang process exit.
- PR #2415 added root-client shutdown, but the custom root loop stops after draining queued messages and does not execute the standard actor teardown path.
- PR #3368 made `shutdown_context()` wait for the root's terminal status and registered shutdown at process exit, but that status still does not imply that the root's event-loop thread has exited.
- PR #4244 added loop cancellation and `loop.stop()` to normal actor cleanup, but the root custom loop does not call that cleanup and normal cleanup does not wait for the thread after scheduling the stop.
- PR #4242 made actor-local event loops unconditional for `PythonActor`.
## Expected Behavior
when an actor reports terminal status, the event-loop thread owned by that actor has exited or cleanup has reached its documented bound. when `shutdown_context().get()` completes, the root actor and actors linked beneath it have completed the same teardown required of normally driven actors.
## Reproduction
the failure reproduces intermittently when the actor-error suite is constrained to two CPUs:
```bash
taskset -c 0,1 pytest -q python/tests/test_actor_error.py
```
the native stack contains CPython thread-finalization frames followed by PyO3/Rust unwind cleanup, and the process exits with `SIGABRT`. the individual test named by pytest usually passes when run alone.
## Proposed Direction
- retain each actor's event-loop thread with the `PythonActor` that created it;
- make bounded actor cleanup wait for that thread after cancelling pending tasks and stopping the loop;
[...]

PR #4757, opened 2026-08-26
stop actor event loops before Python exits (#4757)

shayne-fletcher:
Summary:

GitHub issue: https://github.com/meta-pytorch/monarch/issues/4749
GitHub pull request: https://github.com/meta-pytorch/monarch/pull/4757

each `PythonActor` instance has its own asyncio event loop running on a dedicated daemon OS thread, created through Python's `threading.Thread`. today Python can begin exiting while one of those threads is still alive, which can crash the whole process. specifically, CPython can force an actor thread to exit while it still has Rust/PyO3 calls on its stack. that forced unwind aborts the process with `FATAL: exception not rethrown`.

cpuhrsch's core dumps establish this exact failure path, and D117279357 identified and field-tested the key mitigation: stop actor event loops before Python finalizes. this diff builds on that idea and refines how the loops are found and stopped:

- each actor-loop thread is marked with its loop before it starts, allowing shutdown to use Python's existing list of live threads instead of adding another global registry or lock;
- the existing `shutdown_tokio_runtime` handler performs the work, so no second production `atexit` handler is added;
- shutdown asks every marked loop to stop, releases the GIL so the threads can run, and polls Python's live-thread list until the shared deadline. this keeps shutdown bounded even on Python 3.14, where `Thread.is_alive()` can itself block;
- every loop shares one 250 ms deadline instead of receiving its own timeout. the actual bound is 250 ms plus one final scan.

the guard is deliberately best effort. it covers loops visible while it is scanning, but it can miss a loop created after it has scanned and found none.

together with this fix, a deterministic subprocess test is added that follows the real shutdown path and reproduces precisely the observed failure: without the markers, Python 3.12 aborts during shutdown; with them, every actor loop stops before exit on Python 3.12 and GIL-enabled Python 3.14. focused Rust tests cover multiple loops and the deadline path.

this is a temporary mitigation. it stops leftover actor loops just before Python exits, but it does not fix why they are still running. today actor cleanup can return before its loop thread ends, and the root client can report `Stopped` without running that cleanup. the pytokio-removal work will provide the lifecycle fix. once every actor owns and waits for its loop thread on every shutdown path, this mitigation and its marker can be removed.

the Jobs API creates a simple process tree: it provisions worker host processes with `run_worker_loop_forever()`, and those worker hosts launch actor processes through `bootstrap_main`. a worker host hands control to Rust and ends with `process::exit(0)`, so it never runs Python finalization. the client and actor processes return through Python and do run `Py_FinalizeEx`; those are the processes this mitigation protects.

Differential Revision: D117568016

### DIFF
--- a/monarch_hyperactor/src/actor.rs
+++ b/monarch_hyperactor/src/actor.rs
@@ -84,6 +84,7 @@ use crate::pympsc;
 use crate::pytokio::PyPythonTask;
 use crate::runtime::GilSite;
 use crate::runtime::get_tokio_runtime;
+use crate::runtime::mark_actor_event_loop_thread;
 use crate::runtime::monarch_with_gil;
 use crate::runtime::monarch_with_gil_blocking;
 use crate::supervision::PyMeshFailure;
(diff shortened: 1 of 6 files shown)


shayne-fletcher:
Fixes https://github.com/meta-pytorch/monarch/issues/4749

